PR #732, opened 2026-05-29
fix: repair the CI test suite (conda.sh lookup + --env-dev flag)

cailmdaley:
Turning on the `develop` CI trigger surfaced a chain of pre-existing failures in the test suite that had been invisible (CI only ran on PRs to `main`/`master`). This PR fixes the two that block the suite from running.

## 1. `install_shapepipe` couldn't find `conda.sh`

It only looked under `/opt/conda` (docker convention) or `$CONDA_PREFIX`, neither of which resolves on GitHub runners (`$CONDA_PREFIX` is empty in the install step's shell). Fixed by querying conda itself via `conda info --base`, with the old paths kept as fallbacks. Verified locally against a non-standard prefix (`/softs/intelpython/...`) the old logic would miss.

## 2. CI built the wrong conda environment

The install steps called `./install_shapepipe --develop`. But `--develop` only adds dev *pip* extras into the **production** env; the flag that builds the dev conda env (`shapepipe-dev`, from `environment-dev.yml`) is `--env-dev` (which also triggers `.[dev]` on its own). The downstream `Run tests` step already does `conda activate shapepipe-dev` — an env the old flag never created. Switched both install steps to `--env-dev`.

## Out of scope — separate production bug

`environment.yml` (production) is itself unsolvable: `astropy==5.2` (py38–py311 builds only) pinned against `python=3.12`. This breaks real production installs and deserves its own issue/PR; it's not on the dev-env path the test suite uses.

— Claude on behalf of Cail

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -52,12 +52,12 @@ jobs:
           CC: gcc-9
           CXX: g++-9
         if: runner.os == 'Linux'
-        run: ./install_shapepipe --develop
+        run: ./install_shapepipe --env-dev
 
       - name: Install package (macOS)
         shell: bash -l {0}
         if: runner.os == 'macOS'
-        run: ./install_shapepipe --develop
+        run: ./install_shapepipe --env-dev
 
       - name: Run tests
         shell: bash -l {0}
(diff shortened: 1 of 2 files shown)


cailmdaley:
Superseding this. The deeper fix is to remove conda from CI entirely: turning on the develop trigger showed the conda test path had rotted (this PR's conda.sh + `--env-dev` fixes were layers 1–2 of an N-layer install failure), but the Docker work already moved the runtime to a slim-Python + apt + uv-lockfile story with no conda at all. Rather than repair a parallel conda env nothing deploys, CI will build the dev image and run pytest inside it, and `install_shapepipe` + `environment*.yml` will be deleted. Tracked in the conda-removal PR to follow.

— Claude on behalf of Cail

PR #725, opened 2026-05-10
Fix centroid shift

aguinot:
## Summary

Fix the unit of the centroid guess and (row; col) / (x; y) order. Maybe there is something better to do than setting it to 0 incase the guess fails.
Also fix the size type as `get_noise` expect `T` instead of `sigma`. This is not a huge bug because the size was overestimated.
(Fix a small typo)

## Reviewer Checklist

- [ ] The PR targets the `develop` branch
- [ ] The PR is assigned to the developer
- [ ] The PR has appropriate labels
- [ ] The PR is included in appropriate projects and/or milestones
- [ ] The PR includes a clear description of the proposed changes
- [ ] If the PR addresses an open issue the description includes "closes #<ISSUE NUMBER>"
- [ ] The code and documentation style match the current standards
- [ ] Documentation has been added/updated consistently with the code
- [ ] All CI tests are passing
- [ ] API docs have been built and checked at least once (if relevant)
- [ ] All changed files have been checked and comments provided to the developer
- [ ] All of the reviewer's comments have been satisfactorily addressed by the developer

Issue #656, opened 2023-10-11
[BUG] WCS ngmix bug


lbaumo:
There is a problem with the WCS in ngmix.

lbaumo:
see the ngmix issue [here](https://github.com/esheldon/ngmix/issues/72)

Issue #604, opened 2023-01-11
[NEW FEATURE] Weight Handling


lbaumo:
Ngmix requires variance maps (or inverse variance maps, I can't ever remember which) as input.  Theli weights need to be rescaled by the inverse sky variance.  Megapipe weights are mysterious, so non-zero pixels are set to 1 then scaled by the inverse background variance.  This inverse variance is currently calculated per postage stamp.  I think this should be updated to the sextractor value (which should be already present in the psfex run) in the vignet step.  Also, we need to find out what the megapipe weights actually contain.

lbaumo:
I found [this](https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/en/megapipe/docs/desc.html) "The values of a weight map pixel is proportional to the inverse of the variance of that pixel in the main image."  This documentation is for an older version of megapipe when cfht had 36 ccds.  However, the images themselves seem to have pixel values of 1 or zero.  Time to email Stephen Gwyn to make sure.

lbaumo:
from Stephen Gwyn "The weight images for the individual images are, strictly speaking, masks (0 or 1)"

aguinot:
What ngmix needs are inverse variance weights.
Those maps can almost be created for free with the pipeline as it is. The procedure that is used to get the background images can be applied to create an rms map. Basically by just adding `BACKGROUND_RMS` to the list of CHECKIMAGE_TYPE. Then, just like we do with the background, the pipeline can creates postage stamps with the vignet_maker at galaxy positions. Finally there will be a little bit of work in the ngmix_runner to read those files as input and compute the `1/(BACKGROUND_RMS)^2` which should be very easy to do.
   
Side note: I just noticed that SExtractor can actually produce a background subtracted image by adding `-BACKGROUND` to the list of check images. We could consider doing that instead of computing it in the ngmix_runner. Results should be identical though.

martinkilbinger:
Would be good to use the THELI weight images.

Issue #198, opened 2020-05-18
Mask axel - [merged]


sfarrens:
In GitLab by @aguinot on Jan 21, 2019, 16:52

_Merges mask_Axel -> master_

Implementation of Mask module.

PR #741, opened 2026-06-01
Ngmix v2.0 (CI mirror of #740)

### LINKED ISSUE #762: ngmix v2.0 — a correct, validated v2.0 shape catalogue
**Tracking issue (epic)** for the ngmix v2.0 shape-measurement upgrade — the keystone of the UNIONS tomographic round. PRs merge and fall away; this issue is the durable home for the *goal*, and the single place to see every moving piece, across both `shapepipe` and `sp_validation`.

## The goal

A v2.0 shape catalogue we trust end-to-end: ngmix 2.4.0 + Lucy Baumont's new ngmix classes, PSF/shape columns that mean what their names say, the known shape-measurement bugs fixed, and the downstream validation consuming the new columns.

## Pieces

**Shape measurement — `shapepipe`**
- #741 — Ngmix v2.0 migration (the keystone PR)
- #761 — PSF columns: single source of truth (original vs reconvolved PSF)
- #749 — PSF-prior bug: the PSF fit is prior-dominated → biased PSF columns *(tracked as a sub-issue below)*

**Downstream — `sp_validation` (cross-repo)**
- CosmoStat/sp_validation#201 — adopt the renamed v2.0 PSF/shape columns. *The column rename in #761 forces a matching change in validation, so this lives in the other repo but is part of the same initiative.*

## Not in this epic

- **General ShapePipe test guardrails** (star shear-response, m-bias) — these test ShapePipe broadly, not v2.0 specifically; tracked separately.
- **ngmix config knobs** (`metacal_psf` / `metacal_step` / `galaxy_model`) — deferred; not on the board yet.

— Claude on behalf of Cail

cailmdaley:
## What this is

A same-repo mirror of **#740 (@martinkilbinger's "Ngmix v2.0")**, pushed to a branch on `CosmoStat/shapepipe` so that **CI actually runs**. All 57 commits are authored by Martin (and carry Lucy Baumont's and Axel Guinot's work) — pushing the branch preserves that authorship unchanged; the only thing this PR adds is a same-repo head so GitHub Actions fires the `pull_request` workflow without the fork-PR approval gate. #740 received no CI runs at all for this reason.

**Substance is identical to #740** — see that PR for the full description. In short: upgrade ngmix to 2.4.0 and adopt Lucy's new ngmix classes/interface; overhaul the shape-measurement module; centroid-bias fix + validation; v2.0 production-run plumbing.

Going forward, opening PRs directly on `CosmoStat/shapepipe` (rather than from a fork) avoids this — fork PRs don't trigger our Docker-image CI without a maintainer approval that wasn't happening.

Closes/supersedes #740 once CI is green (leaving that call to Martin).

## Review

A detailed review is on its way (read against Martin's checklist plus a science-quality pass). Headline from exercising the new fitter against real ngmix 2.4.0 on candide: the metacal path runs end-to-end and shear recovery is unbiased at the few×10⁻⁴ level in m. Full notes to follow.

— Claude on behalf of Cail


src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -950,241 +920,248 @@ def get_noise(gal, weight, guess, pixel_scale, thresh=1.2):
...
-        "g_PSFo": np.array([0.0, 0.0]),
-        "g_err_PSFo": np.array([0.0, 0.0]),
-        "T_PSFo": 0.0,
-        "T_err_PSFo": 0.0,
-    }
-    gal_guess = []
-    gal_guess_flag = True
-    wsum = 0
-    for n_e in range(n_epoch):
+    noise_img = rng.standard_normal(gal.shape) * sig_noise
```
cailmdaley, line 1043: **Reproducibility fix (done in this branch).** The module sets `self._rng = RandomState(seed)` per tile precisely so a rerun reproduces — but `prepare_ngmix_weights` was drawing the noise image and the masked-pixel fill from the *unseeded* global `np.random.randn`. Net effect: the same tile with the same seed produced different shears (observed `max|Δg1| ≈ 3e-5` between two identical-seed runs). Fixed by threading `rng` through `make_ngmix_observation → prepare_ngmix_weights` and drawing via `rng.standard_normal`. Guarded by the new `test_metacal_is_reproducible_with_fixed_seed` — fails before this change, passes after.

— Claude on behalf of Cail

tests/module/test_ngmix.py
```diff
@@ -47,3 +47,50 @@ def test_megacam_flip_leaves_unrotated_ccds_unchanged(rows, ccd_nb):
...
+        psfs,
+        weights,
+        flags,
+        jacobs,
+    )
+    res, _ = do_ngmix_metacal(stamp, prior, 1.0, rng)
+    return np.asarray(res["noshear"]["g"])
+
+
+def test_metacal_is_reproducible_with_fixed_seed():
```
cailmdaley, line 126: New property test for the reproducibility fix: same seed → identical noshear `g`. Red on the pre-fix code (global-RNG leak), green after threading `rng`. Kept cheap (51 px, 2 epochs) so it runs comfortably in the suite.

tests/unit/test_scripts_import.py
```diff
@@ -0,0 +1,44 @@
...
+# excluded by deletion: it targets the old test_centroid_bug branch/env (run via
+# run_bias_test.sh's SHAPEPIPE_PATH), not this module.
+VALIDATION_SCRIPTS = [
+    "scripts/validation/centroid/centroid_bias_v2.py",
+    "scripts/python/fitting.py",
+]
+
+
+@pytest.mark.parametrize("relpath", VALIDATION_SCRIPTS)
+def test_validation_script_imports_cleanly(relpath):
```
cailmdaley, line 33: **New smoke test + a removal it triggered.** `tests/unit/test_imports.py` walks the `shapepipe` *package*, but standalone `scripts/` files live outside it and were never import-checked. This extends import-smoke into `scripts/`, and it immediately caught that the v1 `scripts/validation/centroid/centroid_bias.py` imported `get_guess` — a helper the v2.0 rewrite removed (v2.0 seeds fits via ngmix's own guessers; the centroid piece survives inline in `make_ngmix_observation`). That v1 script is the old-API bias repro and is only ever run against the `test_centroid_bug` worktree (`run_bias_test.sh` invokes `${SHAPEPIPE_PATH}/${TEST_SCRIPT}`), so the copy on this branch was dead and only broke import. We removed the stranded copy here; `centroid_bias_v2.py` supersedes it for v2.0. If you'd rather keep a v1 copy on this branch, it'd need `get_guess` vendored in.

— Claude on behalf of Cail

cailmdaley (commented):
Review of the v2.0 work — **part 1 of 2**, scoped to the few cut-and-dry items we could fix and *test* outright. I've committed these here on the CI mirror so the suite proves them (green: 260 passed); inline notes point at each.

A second pass with the more substantive, science-level questions (the `*_psfo` columns now reporting the metacal-reconvolved PSF rather than the original PSFEx/MCCD PSF; the weight-map normalization change; the `r50`/`T` column units; a few others) will follow once we've talked them through — those are judgement calls I don't want to assert unilaterally.

Context for this mirror PR is in the description above. Drafted by Claude, directed by me through the review.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -85,18 +229,29 @@ def __init__(
...
+
+      
 
         self._output_dir = output_dir
         self._file_number_string = file_number_string
 
         self._zero_point = zero_point
         self._pixel_scale = pixel_scale
 
+        # MKDEBUG check whether still used
```
cailmdaley: **`CHECK_EXISTING_DIR` resume looks silently dropped.** This `# MKDEBUG check whether still used` is on the money: the new `process()` never reads `self._check_existing_dir` and never calls `get_last_id` (both now dead). So a configured resume reprocesses from scratch and overwrites, rather than skipping already-done objects — a documented feature that now no-ops. Either rewire the resume path or remove the option + `get_last_id` so it can't mislead. (And the `# MKDEBUG` marker can go.)

src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -269,27 +401,30 @@ def compile_results(self, results):
...
                 )
                 output_dict[name]["T"].append(results[idx][name]["T"])
                 output_dict[name]["T_err"].append(results[idx][name]["T_err"])
-                output_dict[name]["Tpsf"].append(results[idx][name]["Tpsf"])
-                output_dict[name]["g1_psf"].append(
-                    results[idx][name]["gpsf"][0]
+                output_dict[name]["Tpsf"].append(results[idx]["T_PSFo"])
+                output_dict[name]["g1_psf"].append(results[idx]["g_PSFo"][0])
+                output_dict[name]["g2_psf"].append(results[idx]["g_PSFo"][1])
+                output_dict[name]['r50'].append(results[idx][name]['pars'][4])
```
cailmdaley: **`r50` column naming.** For a `gauss` model `pars[4]` is `T` (≈2σ², arcsec²) — the same value already stored under `['T']` a few lines down. And `r50_PSFo = sqrt(T_psf/2)` (line ~676) is the Gaussian σ, not the half-light radius (true `r50 = 1.1774·σ`). sp_validation consumes these size columns downstream, so it matters what they actually contain. Is `r50` intentional shorthand for "a size," or should these be genuine half-light radii?

src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -462,422 +597,258 @@ def process(self):
...
+    for expccd_name in psf_expccd_names:
+        exp_name, ccd_n = re.split('-', expccd_name)
 
-    hsm_shape = galsim.hsm.FindAdaptiveMom(galsim_img, strict=False)
+        gal_vign = (
+            vignet.gal_vign_cat[str(obj_id)][expccd_name]['VIGNET']
+        )
 
-    error_msg = hsm_shape.error_message
+        if np.all(gal_vign == 0):
```
cailmdaley, line 784: **Zero-pixel guard loosened `any`→`all`.** v1 skipped an epoch if *any* pixel was exactly 0; this now skips only if `np.all(gal_vign == 0)`. Partial CCD-edge stamps (a band of exact-zero pixels) will now enter the fit and rely on flags/weights to catch them. Intended loosening, or should the partial-zero case still be excluded?

src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -462,422 +597,258 @@ def process(self):
...
-            f"invalid guess_flux_unit '{guess_flux_unit}',"
-            + " must be one of 'img', 'sky'"
+        flag_vign = (
+            vignet.flag_vign_cat[str(obj_id)][expccd_name]['VIGNET']
         )
+        if tile_vign is not None:
+            flag_vign[np.where(tile_vign == -1e30)] = 2**10
+        v_flag_tmp = flag_vign.ravel()
+        # remove objects that are more than 1/3 masked
+        if len(np.where(v_flag_tmp != 0)[0]) / (51 * 51) > 1 / 3.0:
```
cailmdaley: **Hardcoded `51 * 51` denominator.** The masked-fraction cut divides by a literal `51 * 51`. If the stamp size ever becomes configurable (which is on the near-term roadmap) this silently goes wrong. `v_flag_tmp.size` would track the actual stamp.

src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -950,241 +920,248 @@ def get_noise(gal, weight, guess, pixel_scale, thresh=1.2):
...
+    numpy.ndarray
+        Noise image.
     """
-    n_epoch = len(gals)
+    weight_map = np.copy(weight)
+    weight_map[flag != 0] = 0.0
 
-    if n_epoch == 0:
-        raise ValueError("0 epoch to process")
+    sig_noise = sigma_mad(gal)
```
cailmdaley: **Weight-map normalization changed vs v1.** The noise estimate `sigma_mad(gal)` here is taken over the whole object-containing stamp (flux-contaminated), where v1 used a windowed, object-free estimate (`get_noise`). And the returned weight keeps the per-pixel (FSCALE-rescaled) map *and* gets multiplied by `1/sig_noise²` — which reads as a double inverse-variance weighting. Between the two, this shifts the fit's χ² weighting, the reported errors/s2n, and the PSF-averaging weights relative to the validated v1. Is this the intended new weighting?

src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -950,241 +920,248 @@ def get_noise(gal, weight, guess, pixel_scale, thresh=1.2):
...
+    names = ['T_psf', 'T_psf_err', 'g_psf', 'g_psf_err']
+    psf_dict = {k: [] for k in names}
+    nepoch = len(obsdict['noshear'])
+    wsum = 0
+    g_psf_sum = np.array([0., 0.])
+    g_psf_err_sum = np.array([0., 0.])
+    T_psf_sum = 0
+    T_psf_err_sum = 0
+    for n_e in np.arange(nepoch):
+        T_psf=obsdict['noshear'][n_e].psf.meta['result']['T']
```
cailmdaley: **`*_psfo` columns now carry the metacal-*reconvolved* PSF.** `obsdict['noshear'][n_e].psf.meta['result']` is the reconvolved metacal PSF, and it feeds `T_PSFo`/`g_PSFo`/`r50_PSFo`. The `compile_results` comment still describes these as "the original image psf from psfex or mccd," and downstream PSF-leakage / ρ-stat / null tests consume them as the *input* PSF. v1 fit the raw pre-metacal PSF separately. Also note this collapses the per-type PSF columns — `Tpsf`/`g1_psf`/`g2_psf` end up identical across all 5 metacal HDUs. Fine if the metacal target PSF is type-independent, but any step that differences them now gets exactly zero. Intended? If so the comment needs updating.

src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -950,241 +920,248 @@ def get_noise(gal, weight, guess, pixel_scale, thresh=1.2):
...
-        "step": 0.01,
-        "psf": "gauss",
-        "fixnoise": True,
-        "cheatnoise": False,
-        "symmetrize_psf": False,
-        "use_noise_image": True,
-    }
 
-    Tguess = np.mean(T_guess_psf)
+def do_ngmix_metacal(stamp, prior, flux_guess, rng):
```
cailmdaley: **PSF fit reuses the galaxy joint prior and the galaxy `FLUX_AUTO` as the PSF flux guess.** This mirrors the upstream ngmix example's prior reuse, but a galaxy-flux guess for the PSF fit is a little unusual and could touch PSF-fit convergence/quality. Lower-confidence flag — worth a sanity check that PSF fits converge as expected.

src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -134,58 +290,28 @@ def MegaCamFlip(self, vign, ccd_nb):
         if ccd_nb < 18 or ccd_nb in [36, 37]:
             # swap x axis so origin is on top-right
             return np.rot90(vign, k=2)
+            print('rotating megapipe image')
```
cailmdaley: Cleanup: this `print('rotating megapipe image')` is after the `return`, so it's unreachable — can be dropped.

src/shapepipe/modules/ngmix_package/ngmix.py
```diff
@@ -950,241 +920,248 @@ def get_noise(gal, weight, guess, pixel_scale, thresh=1.2):
...
-            wsum += twsum
-            gpsf_sum[0] += g1 * twsum
-            gpsf_sum[1] += g2 * twsum
-            Tpsf_sum += T * twsum
-            npsf += 1
+    boot = ngmix.metacal.MetacalBootstrapper(
+        runner=runner,
+        psf_runner=psf_runner,
+        ignore_failed_psf=True,
+        rng=rng,
```
cailmdaley, line 1415: Cleanup: `sextractor_e1e2` looks unused here, and its docstring appears copy-pasted from another function (doesn't describe this one). Remove, or fix the docstring if it's kept.

src/shapepipe/modules/sextractor_runner.py
```diff
@@ -68,6 +68,12 @@ def sextractor_runner(
     else:
         prefix = None
 
+    # When post-processing is enabled the sqlite WCS file is the last input;
+    # remove it before passing the image files to SExtractorCaller.
+    if config.getboolean(module_config_sec, "MAKE_POST_PROCESS"):
+        f_wcs_path = input_file_list[-1]
```
cailmdaley, line 78: **Runner input-contract now positional, but the decorator wasn't updated.** This pulls the WCS path from `input_file_list[-1]` — the WCS moved from a named config option into the positional input list. The same positional move happened in `read_ext_sexcat_runner`, `psfex_interp_runner`, and `vignetmaker_runner`, but (unlike `ngmix_runner`) their `@module_runner` decorators weren't updated to match. The shipped v2.0 example configs are correct, so this isn't a live bug — but the real input contract now lives entirely in each config's `FILE_PATTERN`/`FILE_EXT` ordering, so a hand-written or v1-style config would silently mis-map inputs or `IndexError` with nothing to catch it. `ngmix_runner` is the model — worth extending these four decorators to match so the contract is self-documenting.

cailmdaley (commented):
## Review — part 2 of 2 (methodology + remaining items)

*Provenance: this review was generated by Claude, working on candide against the actual module code; I haven't hand-edited it line by line, so please read it with that lens. The empirical checks below were genuinely run; the inline questions are real intent questions, not assertions that something is broken.*

Part 1 covered the cut-and-dry items we fixed and tested directly (the RNG reproducibility fix + its property test, the import smoke-test and the `get_guess`/`centroid_bias.py` removal it surfaced). This part is the broader read: what we verified empirically, plus methodology and housekeeping. The code-specific points are left as **inline comments** on the relevant lines.

### What we verified empirically ✅

We installed **ngmix 2.4.0** (from this PR's own `esheldon/ngmix@v2.4.0` source) over the container stack and ran the **actual module code**:

- All 13 ngmix-2.4 symbols the module uses resolve; the metacal/bootstrap path runs end-to-end.
- On 200 sub-pixel-shifted galsim galaxies, `do_ngmix_metacal` recovers shear with **m = +2×10⁻⁴ ± 3×10⁻⁴**, **c₂ = −1×10⁻⁵** (both consistent with zero), R₁₁≈R₂₂≈0.924, 200/200 OK. Your `centroid_bias_v2.py` (noiseless) gives m = +3.7×10⁻⁴.
- **The centroid fix** (HSM re-centering of the galaxy Jacobian in `make_ngmix_observation`) runs clean and is benign in the high-S/N regime we tested. We didn't try to reproduce the bias it targets — Cail notes that lives in the old ngmix-1.x path, so current code can't exhibit it. Just confirming the new fix doesn't *introduce* anything.

The 2.x API is used correctly, the fitter is well-behaved, and the CI suite is green on this branch.

### Methodology questions (for you / Lucy)

Five places where v2.0 behaviour differs from the validated v1 in ways that may be intentional but aren't documented in the diff — flagged inline so the choice is explicit:
- weight-map normalization (inline at `prepare_ngmix_weights`)
- `*_psfo` columns now carrying the metacal-reconvolved PSF, and per-type PSF columns collapsing (inline at `average_multiepoch_psf`)
- the galaxy zero-pixel guard loosening `any`→`all` (inline at `prepare_postage_stamps`)
- the PSF fit reusing the galaxy prior + `FLUX_AUTO` flux guess (inline at `do_ngmix_metacal`)

### Checklist / housekeeping

- **`closes #` links** — none on the PR. Axel's centroid-shift PR **#725** overlaps this; worth an explicit cross-ref on whether this supersedes it.
- **Docs** — no `docs/` changes, and the ngmix-pin note in `CLAUDE.md` ("don't modernize this dependency line") is now stale, since this PR *is* the modernization. Worth updating on/with merge.
- **`pyproject.toml`** — ngmix is pinned to a git tag (`esheldon/ngmix@v2.4.0`). If 2.4.0 is on PyPI, `ngmix==2.4.0` would be more reproducible and drop the git dependency — worth a check.
- **Description** — for a +3894/−3410, 67-file PR bumping a core dependency and rewriting the shape module, a fuller description (especially the centroid-bias result) would help future readers.

### A couple of pre-existing cleanups (not changed here, low priority)

In `save_results`: the error string ~`ngmix.py:547` is non-`f`, so it prints literal `{ext_name}`; there's leftover `print(...)` debug just below it; and the `np.append`-per-batch FITS write is O(n²) over a tile. Worth a sweep if you're in there.

Happy to push fixes for any of the cut-and-dry ones (cleanups, the `CHECK_EXISTING_DIR` plumbing, the runner decorators) — just say which.

— Claude on behalf of Cail

scripts/python/fitting.py
```diff
@@ -0,0 +1,265 @@
+"""
```
cailmdaley: **Is this meant to ship?** This reads as a near-verbatim copy of esheldon's `fitting_bd_empsf.py` ngmix example (bulge+disk `fracdev`, `espy`/`images` refs in the docstring) and doesn't look connected to the pipeline. It also redefines its own `make_data`/`get_prior`, duplicating the new `shapepipe.testing.simulate` helpers added in this same PR (whose point is to be the single stable-across-branches source). Suggest dropping it, or trimming it to import the canonical helpers. `scripts/jupyter/test_centroid_shift.py` has the same duplication and could likewise import `simulate.make_data`.

src/shapepipe/modules/ngmix_package/ngmix.py
martinkilbinger: Good. This probably was previous code before we changed to a random seed per tile.

cailmdaley:
## Next steps — triage of the v2.0 review

*Provenance: like the two-part review above, this summary was generated by Claude working on candide against the current branch head; I haven't hand-edited it line by line, so please read it with that lens. The re-review was genuinely run — every finding below was re-checked against the code as it stands on this branch.*

A follow-up now that #740 is closed and everything lives here on #741. I re-read the ngmix module and the four runners against the current head: no commits have landed since the part-2 review, so all eleven line-level findings still stand exactly where they were anchored — nothing was silently addressed or drifted. #741 is green and mergeable; the question is just which findings to land before merge and which can wait.

The useful cut isn't open-vs-resolved (they're all open) — it's **what unblocks each one**:

### A · Cut-and-dry — mechanical, no science (one cleanup commit)
- **`ngmix.py:293`** — unreachable `print` after `return`; delete.
- **`ngmix.py:1140`** — `sextractor_e1e2` is unused with a copy-pasted docstring; drop or fix.
- **`ngmix.py:766`** — masked-fraction cut divides by a literal `51 * 51`; use `v_flag_tmp.size` so it survives a configurable stamp size.
- **`sextractor_runner.py:74` (+ `read_ext_sexcat`, `psfex_interp`, `vignetmaker`)** — WCS moved to a positional input but the `@module_runner` decorators weren't updated to match (only `ngmix_runner` was). Shipped configs are correct so it's not a live bug, but the decorator is the input contract.
- **`scripts/python/fitting.py`** — reads as a verbatim esheldon example duplicating the new `testing/simulate.py`; confirm it's not meant to ship, then drop or trim to import the canonical helpers.

Happy to push these as a single commit here if that's useful — they keep CI green and narrow the review surface to the science.

### B · Decisions — need your / Lucy's intent
Three are light (a one-line "intended" or a small fix): **zero-pixel guard `any`→`all`** (`:737`), **PSF fit reusing the galaxy prior + `FLUX_AUTO` flux guess** (`:1068`), and **the `r50` column naming** (`:415`/`:676` — these carry `T` and `sqrt(T/2)=σ`, not half-light radii; worth reconciling with sp_validation, which consumes them).

Two genuinely change the meaning of output columns and are worth a real answer:
- **Weight-map normalization (`:949`)** — `sigma_mad` is now over the whole (flux-contaminated) stamp rather than the windowed object-free `get_noise`, and the per-pixel map is then multiplied by `1/sig_noise²` (reads as double inverse-variance). This moves the χ²-weighting, the reported errors/s2n, and the PSF-averaging weights.
- **`*_psfo` columns (`:1045`)** — these now carry the metacal-*reconvolved* PSF and collapse the per-type columns to one value, where the ρ-stat / null tests consume them as the *input* PSF (and the column doc still says "original image psf").

### C · Resume feature
- **`CHECK_EXISTING_DIR` (`:254`/`:257`/`:464`)** — `_check_existing_dir` is set but never read and `get_last_id` is never called, so a configured resume silently reprocesses from scratch. Rewire it, or remove the option + `get_last_id` (the latter folds into the Bucket-A commit). The adjacent `# MKDEBUG` sits on `_f_wcs_path`, which *is* used — that marker can just go.

### Suggested sequence to merge
1. Land the Bucket-A cleanups (I can push these).
2. Decide the resume question (rewire vs remove).
3. Confirm the Bucket-B science questions — the two above are the only ones I'd treat as genuine merge-gates; the rest can merge-then-iterate.
4. Reconcile r50/T naming with sp_validation.
5. Merge.

Honest read: CI is green and nothing here breaks the suite, so #741 *can* merge today — but the weight-norm and `*_psfo` changes silently alter columns that sp_validation and the null tests rely on, so those two are worth a conscious accept rather than a default one.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
martinkilbinger: This was a hack to resume interrupted ngmix runs, bookkeeping done in scripts outside modules. Can be removed now, and if needed again in the future implemented better within the module.

src/shapepipe/modules/ngmix_package/ngmix.py
martinkilbinger: The intend is to resolve the existing confusion about 'T', which in the DES papers is the area measure 2 * sigma^2, but in our ngmix implementation was the half-light radius, so the name was confusing. This was set by `guess_size_type="sigma"` in the earlier version in the call to get_guess(), which does not exist anymore.

We need to check that `pars[4]` is indeed 'T', and that `r50` should be `1.1774·σ` and not just `σ`. Would be good to not hard-code these transformations. There exist some functions in `sp_validation:galaxy.py`, we would have to move them to e.g. `cs_util`.

src/shapepipe/modules/ngmix_package/ngmix.py
martinkilbinger: This change was added in commit #d01503d0: the any == 0 test had pass close to no postage stamps. I though this was too restrictive.

src/shapepipe/modules/ngmix_package/ngmix.py
martinkilbinger: yes would be good to not hard-code this.

martinkilbinger: We have in other module's config files the entries
`STAMP_SIZE = 51`
or `VIGNET_SIZE = 51`

src/shapepipe/modules/ngmix_package/ngmix.py
martinkilbinger: Poking @lbaumo

src/shapepipe/modules/ngmix_package/ngmix.py
martinkilbinger: There is this discussion about noise and weight maps: https://github.com/CosmoStat/shapepipe/issues/604

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: The fix is on the branch: 22f1f0a1 threads the per-tile seeded `rng` through `make_ngmix_observation → prepare_ngmix_weights`, so all noise draws now come from the tile seed. `test_metacal_is_reproducible_with_fixed_seed` guards it (red before, green after).

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: Removed as agreed: bd60dc8e drops the whole resume path (`_check_existing_dir`, `get_last_id`, the `# MKDEBUG` markers) from `ngmix.py` + `ngmix_runner.py`, and 16789eb3 drops the now-dangling `CHECK_EXISTING_DIR` entry from `example/cfis/config_tile_Ng_template_batch.ini`. The live `CHECK_EXISTING_DIR` features in `mask`/`get_images` are untouched.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: This now exists end-to-end, in the shape you describe — honest names, no hard-coded transformations, conversions in cs_util:

- **Verified:** `pars[4]` is indeed `T = 2σ²` for the gauss model, and the true half-light radius is `r50 = 1.1774·σ = √(ln2·T)` (the factor is √(2 ln 2); both pinned in tests).
- **cs_util** CosmoStat/cs_util#65 (merged, released as cs_util 0.2.1 on PyPI): new `cs_util.size` module holding the full T ↔ σ ↔ r50 ↔ FWHM conversion web, with tests pinning values and inverse round-trips.
- **shapepipe** #743 (merged into this branch, d6531b1b): every `r50` column is now a genuine half-light radius — galaxy `r50 = √(ln2·T)` with propagated error (NaN for non-positive fitted T), PSF `r50psf = √(2 ln2)·σ` with a corrected error (the previous `T_err/(2σ)` was a factor-2 over-estimate). The duplicate `*_psfo_ngmix` size columns (bit-for-bit copies of `Tpsf`/`r50psf`) are retired and `make_cat` reads `Tpsf`. This matches the UNIONS-3500 WL I convention: r_h primary, T the derived DES area.
- **sp_validation** CosmoStat/sp_validation#198 (open): switches `galaxy.py` to the cs_util converters — and that surfaced a real bug worth your attention: the local `T_to_fwhm` was dimensionally wrong (`T/1.17741·2.355` treats the area T as if it were σ; correct is `2.35482·√(T/2)`). A monotonic but nonlinear distortion of the PSF-leakage size axis, so per-bin α(size) trends made through the old function should be regenerated. #198 is held un-merged only by `shear_psf_leakage`'s `cs-util ^0.1.0` cap — a one-line relaxation on its develop unblocks it (details in the PR body).

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: That makes sense — with `any`, a single dead pixel anywhere in a 51×51 stamp kills the whole epoch, so passing almost nothing is exactly what you'd expect. Agreed it stays as `all`. If partial CCD-edge stamps ever show up in null tests, a fractional threshold (skip when more than some fraction of pixels is zero/flagged) would be the principled middle ground — noting as a possible follow-up, not a request for this PR.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: De-hardcoded in bd60dc8e: the denominator is now `v_flag_tmp.size`, and that was the last literal 51 in the module — all geometry now derives from the actual array shapes. We deliberately didn't add a `STAMP_SIZE` entry to the ngmix config: stamp size is owned upstream by vignetmaker's `STAMP_SIZE`/`VIGNET_SIZE`, so a local entry would be a dead knob that could disagree with the actual vignets. Say the word if you'd still like an explicit option.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: The inverse-variance scheme from #604 is now implemented on this branch — and the double-weighting flagged here is gone in the process: `prepare_ngmix_weights` now returns a single inverse variance, never a retained per-pixel map times `1/σ²`.

- f466c987: optional `BKG_RMS_VIGNET_PATH` (the `background_rms_vignet` sqlite from `vignetmaker_runner`); when supplied, unmasked pixels get per-pixel `1/(Fscale·RMS)²`, with weight/flag gating and non-finite/non-positive RMS pixels masked out (6ddea5b2 restores the binarized gating mask).
- Fallback when the option is absent: uniform `mask/σ_MAD²`, hardened in 4aa3b2a1 so a degenerate constant stamp (σ_MAD = 0) yields zero weights instead of NaN.
- 158986a4 pins the chain end-to-end: spatially varying heteroscedastic noise injected via `make_data`, asserting the ngmix `Observation` weight equals `1/(Fscale·rms)²` exactly through `rescale_epoch_fluxes → prepare_ngmix_weights → make_ngmix_observation`. f79de46d covers the optional-input wiring; 0bc6016e documents the all-or-nothing semantics in the four example configs (missing vignet file → `FileNotFoundError`; scalar fallback only when the option is absent).

Would value your eyes on whether this matches what you had in mind in #604.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: Partial movement via #743 (d6531b1b): the duplicate `T_psfo_ngmix`/`r50_psfo_ngmix` columns (bit-for-bit copies of `Tpsf`/`r50psf`) are retired, so the duplication is no longer multiplied across column families. The substantive question stands and is yours to call: these columns carry the metacal-*reconvolved* PSF while the `compile_results` comment (and downstream leakage/ρ-stat consumers) treat them as the psfex/mccd input PSF. Our read: either fit the raw pre-metacal PSF separately as v1 did, or keep the reconvolved one and rename/document it as the metacal target PSF — the only untenable option is the current mismatch between contents and label.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: Still open from our side — no fix attempted, since it's a methodology call. Our current read: likely benign (the prior is broad and the flux guess only seeds the optimizer), but cheap to settle empirically — PSF-fit convergence/failure rates on one real tile, galaxy-flux guess vs. a PSF-appropriate one. Leaving it with you.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: Done in bd60dc8e.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: Removed in bd60dc8e — confirmed no callers anywhere in `src/`.

— Claude on behalf of Cail

src/shapepipe/modules/sextractor_runner.py
cailmdaley: `sextractor_runner` is fixed: d4ada3d2 updates its `@module_runner` decorator to the real positional contract (the `log_exp_headers` `.sqlite` from `merge_headers_runner`, same convention as `ngmix_runner`) and replaces the stale `LOG_WCS` docs in `sextractor_package/__init__.py`; b2dcd793 fixes the same doc drift in `ngmix_package/__init__.py` (the headers file is `input_file_list[6]`, never a config option). We audited all 12 shipped `SEXTRACTOR_RUNNER` config sections — every one overrides `FILE_PATTERN` explicitly, so the decorator-default change alters no behavior. The other three runners (`read_ext_sexcat`, `psfex_interp`, `vignetmaker`) are left as-is for now: their inputs are mode-conditional (e.g. `psfex_interp` ME mode takes `f_wcs_path` as `input_file_list[1]`), so a static decorator default can't express the contract — those want a small follow-up of their own.

— Claude on behalf of Cail

scripts/python/fitting.py
cailmdaley: Dropped in bd60dc8e (no callers, duplicated `testing/simulate.py`); the validation-script smoke list was pruned to match in 0df6bf69. `scripts/jupyter/test_centroid_shift.py` still carries its own `make_data` — left as a possible follow-up rather than touching the notebook-support script in this round.

— Claude on behalf of Cail

cailmdaley:
**Review-round summary.** Every open thread above now has a reply; here's the shape of what landed on `ngmix_v2.0` since the review, and what still awaits your judgment.

**On this branch:**
- **Reproducibility** — seeded `rng` threaded through the noise draws (22f1f0a1), guarded by `test_metacal_is_reproducible_with_fixed_seed`.
- **Dead-code cleanup you greenlit** (bd60dc8e): `CHECK_EXISTING_DIR` resume path + `get_last_id`, the unreachable `print`, unused `sextractor_e1e2`, `scripts/python/fitting.py`, and the `51*51` literal (now `v_flag_tmp.size`). Dangling config entry pruned in 16789eb3.
- **Inverse-variance weights per #604** (f466c987 + 6ddea5b2, 4aa3b2a1, 158986a4, f79de46d, 0bc6016e): optional `BKG_RMS_VIGNET_PATH` gives per-pixel `1/(Fscale·RMS)²` with gating masks; scalar `σ_MAD` fallback hardened; the chain pinned end-to-end by an integration test. The double-weighting flagged in review is gone.
- **Runner contract** (d4ada3d2, b2dcd793): `sextractor_runner`'s decorator now declares the positional headers input; stale `LOG_WCS` docs fixed in both packages. No behavior change in any shipped config (all 12 override `FILE_PATTERN`).

**Companion PRs (the r50/T thread):**
- CosmoStat/cs_util#65 — `cs_util.size` conversion web (T ↔ σ ↔ r50 ↔ FWHM), merged and released as **cs_util 0.2.1** on PyPI.
- #743 — honest half-light radii in every `r50` column per the UNIONS convention, merged into this branch (d6531b1b).
- CosmoStat/sp_validation#198 — adopts the cs_util converters and fixes a dimensionally wrong `T_to_fwhm` that distorted the PSF-leakage size axis (results through the old function should be regenerated). Open, blocked only by `shear_psf_leakage`'s `cs-util ^0.1.0` cap.

**Awaiting your call** (methodology, no code pushed): the `*_psfo` columns carrying the metacal-reconvolved PSF vs. the documented psfex/mccd input PSF; the galaxy prior/`FLUX_AUTO` guess reused for the PSF fit; and whether you want an explicit `STAMP_SIZE` option despite vignetmaker owning the geometry upstream. The zero-pixel `any→all` change stays as you intended; a fractional-bad-pixel threshold is noted as a possible follow-up.

— Claude on behalf of Cail

cailmdaley:
## Review — part 3 (fresh pass)

*Provenance: same convention as parts 1–2 — this is a fresh full-diff pass by Claude working on candide, against head b2dcd793. Every finding below was empirically demonstrated before being fixed, each fix carries a regression test confirmed red on the unfixed code, and the chain is now pushed to this branch (`b2dcd793..05f1584e`). I haven't hand-edited it line by line, so please read with that lens.*

### Blockers (2)

Both live in the v2.0 rewrite's tile flow, and both survive a green suite and an easy-object smoke run — which is exactly why they hid: one sits in a path no test exercised, the other only fires when a fit fails. The fixes restore what we read as the v1 contract, but they deserve your sanity-check of intent, Martin.

1. **`TILE_ID` silently truncates the post-process CCD scan** (`sextractor_script.py`, `make_post_process` — fixed in fcd117fc). The tile-level `merge_headers` (new on this branch) writes `TILE_ID` as the first key of `log_exp_headers<tile>.sqlite`; `make_post_process` derived `n_hdu = len(f_wcs[key_list[0]])` — i.e. the *length of the tile ID string* instead of the CCD count. Every epoch on CCDs ≥ `len(tile_string)` silently vanishes from `N_EPOCH`/`EPOCH_*`, and from there out of ME vignets and shape measurement. Repro: tile `"51"` (len 2), 3 CCDs → an object on CCD 2 comes back `N_EPOCH = 0` instead of 2. The fix filters the metadata key and sizes the scan from a real exposure entry; we audited every other `log_exp_headers` consumer (psfex_interp, vignetmaker, ngmix, mccd interp) — all do keyed `[exp][ccd]` access, so `make_post_process` was the only affected site, and it covers both its callers.

2. **`compile_results` KeyErrors on any failed fit type, discarding the whole tile at save** (`ngmix.py` — fixed in 450f3c19). Re-verified against real ngmix 2.4.0: a failed `run_fitter` *returns normally* with `flags=512` and keys `[errmsg, flags, ier, model, nfev, pars, pars_cov, pars_err]` — none of `g/g_cov/T/T_err/flux/flux_err/s2n`. The old `results[idx][name]["flux"]` access then KeyErrors, and because it happens at batch/final save, one hard object throws away every measured object in the tile. v1 never saw this because ngmix 1.x *raised* on failure, which the per-object try/except caught. Now failed types are recorded with their flags and NaN-filled measurement columns (a `flags == 0` result missing `s2n` still KeyErrors, deliberately).

### Should-fixes (pushed; please confirm intent)

- **`mcal_flags` was hardcoded to `0`** (7a1b3b16). `process()` set `res['mcal_flags'] = 0` unconditionally, so the `NGMIX_MCAL_FLAGS` column `make_cat` writes was constant zero and any `mcal_flags == 0` quality cut passed everything. Restored to the v1 contract — bitwise OR of the per-type fit flags (new `get_mcal_flags` helper). Flagging explicitly: confirm this is the intended semantics for that column.
- **Failed-PSF epochs crashed (or would have contaminated) the multi-epoch PSF average** (00e9f898). With `ignore_failed_psf=True` a failed-PSF epoch stays in the obs dict carrying only `flags`/`pars`; `average_multiepoch_psf` read its `['T']` → KeyError, dropping the object. It now skips `flags != 0` epochs, and `n_epoch_model` is again the count of epochs that survived the PSF fit (the v1 meaning) rather than the number submitted.
- **Config drift, tutorial-facing** (3eb6a66f). The four `example/cfis_simu` configs driven by `job_sp_simu.bash` still used the removed `LOG_WCS`/`ME_LOG_WCS` options; they're migrated to the positional WCS-headers input contract, mirroring the migrated `example/cfis` equivalents. Same commit renames `run_sp_exp_Mh` → `run_sp_tile_Mh_exp` in five `example/cfis` configs (the run name `config_tile_Mh_exp.ini` actually produces) and adds explicit `FILE_EXT` where a 3-entry `FILE_PATTERN` override would trip the new 4-entry decorator default's length check at startup. Related sweep worth doing in this PR: `example/cfis/config_exp_psfex.ini` has the same latent `FILE_PATTERN`/`FILE_EXT` length mismatch.

### Noted, not changed

- **`moments_fail` semantic drift** (document-only): in v1 the column counted moments-initial-guess (`get_guess`) failures; on this branch it counts metacal types with nonzero fit flags. Same name, different meaning — flagged so downstream consumers know.
- **`ntry_fit` → `nfev_fit`** (436bcc8a): the value written is now ngmix-2.x's `nfev` (function evaluations), not the v1 retry count (1–5), so the old name was misleading. We verified the column is unconsumed downstream (`make_cat` never reads it; no reference anywhere in sp_validation), hence an honest rename rather than a compatibility shim. A small dead-code sweep rode along (05f1584e): the orphaned `copyfile` import and the never-read `Tile_cat.size/.e/.theta`; `get_noise` was kept — it has a live caller in `scripts/jupyter/test_centroid_shift.py`.

### Branch state, empirically

With the chain applied the container suite is **270 passed / 1 failed** — the one failure is the known `test_metacal_is_reproducible_with_fixed_seed`, which needs ngmix 2.x the test sandbox doesn't have. Under real ngmix 2.4.0 we ran `do_ngmix_metacal` directly: seed-reproducibility holds, and an injected shear is recovered at **g1 = 0.01996 vs 0.02** after response correction.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: One science addendum now that the inverse-variance scheme from #604 is in: the metacal `fixnoise` noise image is drawn **homoscedastic** while the weights are per-pixel. In `prepare_ngmix_weights`, `weight_map[mask] = 1 / bkg_rms[mask]**2` is per-pixel, but `noise_img = rng.standard_normal(gal.shape) * sig_noise` with `sig_noise = median(bkg_rms[mask])` — so the noise realization metacal uses to cancel correlated noise doesn't carry the stamp's spatial noise structure, while the χ² weighting does. Within a 51-px stamp the RMS variation is small, so this is likely second-order; but if you want them consistent, the per-pixel version is one line — `noise_img = rng.standard_normal(gal.shape) * bkg_rms`, with a fallback to `sig_noise` for invalid pixels. Deliberately **not** implemented in the fix chain — it changes the noise model, so it's your call.

— Claude on behalf of Cail

src/shapepipe/modules/ngmix_package/ngmix.py
cailmdaley: A sharper version of this point, relevant for tomorrow's tutorial discussion: with `psf='fitgauss'`, the reconvolved metacal PSF is **round by construction** — so `g1_psfo_ngmix`/`g2_psfo_ngmix` are identically zero for every object, not just "a different PSF than v1". Any downstream PSF-leakage regression of `e_gal` against these columns is degenerate (zero regressor), and a ρ-stat or null test consuming them silently tests nothing. Worth deciding explicitly whether to (a) keep them as a sanity check that fitgauss did its job, (b) rename so nobody mistakes them for the input PSF (that role is served by the PSFEx/MCCD columns), or (c) drop them. `T_psfo` remains meaningful (the reconvolved PSF size); it's only the shape components that are degenerate.

— Claude on behalf of Cail

martinkilbinger:
My response to the review:

# Blockers:
1. TILE_ID truncation of exposure list. I confirmed, only 7 HDUs, one for each contributing [exposure-ccd], was added to the SExtractor output in post-processing. Good catch and fix.
2. Good fix.

# Should-fixes
- mcal_flags: yes, bitwise-or is what is suppoed to be done
- failed PSF epoch: Fix looks good.
- config files: the obsolete `example/cfis_simu` and `example/cfis/config_tile_MiViSmVi.ini` are now removed
- example/cfis/config_exp_psfex.ini: where is the problem exactly?

# Noted
- moments_fail: where is this being flagged?
- ntry_fit → nfev_fit: sounds good.

cailmdaley:
Replies point by point:

**Blockers** — thanks for the empirical confirmation on the TILE_ID truncation; matches what we traced this afternoon (the cap is `len("<tile_id>") = 7` applied to the per-exposure CCD scan in `make_post_process`; the regression test pins an object on the last CCD keeping all epochs).

**mcal_flags** — intent confirmed, nothing further to do.

**Config files** — `example/cfis_simu` actually arrived deleted via the develop merge I pushed (061ca3c7 — that merge also resolves the PR's conflicts: `uv.lock` regenerated from the merged pyproject, ngmix pin at upstream v2.4.0 preserved, the #742 dependabot bumps absorbed). Your `config_tile_MiViSmVi.ini` deletion rode in cleanly under it.

**`config_exp_psfex.ini` — where exactly:** `[SEXTRACTOR_RUNNER]` overrides `FILE_PATTERN = image, weight, pipeline_flag` (3 entries) with no `FILE_EXT`, so it falls back on the decorator's new 4-entry default (`[".fits", ".fits", ".fits", ".sqlite"]` since WCS headers became a positional input) — the pattern/ext length check then fails at startup. Fixed in 6f4901b8 with an explicit `FILE_EXT = .fits, .fits, .fits`, same treatment 3eb6a66f gave the other configs.

**`moments_fail` — where it's flagged:** only in the review text, no code marker — sorry for the ambiguous wording. The state: it's computed at `ngmix.py:669` as the count of metacal types with nonzero fit flags, and consumed at `make_cat.py:358` into the final catalogue. In v1 the same column counted moments-*initial-guess* failures — so the name no longer matches the meaning, and unlike `ntry_fit` it *does* reach the catalogue. Two options: (a) keep the name and document the new meaning, or (b) rename through `ngmix` + `make_cat` (e.g. `mcal_types_fail`). Your call — happy to do either.

**Incoming on this branch:** a GalSim noise-injection validation of the background-RMS inverse-variance weights (per this afternoon's discussion) — known profile + known per-pixel noise, accuracy and pull-calibration checks, with a binary-weights control under spatially-varying noise so the test fails if the weight wiring were wrong. Numbers will be posted here when it lands.

— Claude (Fable), on behalf of Cail

cailmdaley:
The GalSim noise-injection validation has landed (`c5a9cc4d`, tightened in `6bae2a41`) — and it earned its keep before arriving: **it caught a real defect in the rms-weight path as previously merged** (fixed in `bb93c288`, please review that one specifically).

**The finding.** In `do_ngmix_metacal`, the fixnoise noise image was drawn at a *scalar* `median(rms)`. Under a strong noise gradient (factor-8 across the stamp) that floods the low-noise half with excess noise, and the ivar weights were measurably **counterproductive** end-to-end: shape scatter 0.116 (ivar) vs 0.078 (binary), ratio 1.48. Drawing the noise image per-pixel from the rms map restores the expected advantage: 0.053 vs 0.078 (ratio 0.684). Constant rms maps produce bit-identical draws to the old path, and `bkg_rms=None` is untouched.

**What the test pins** (N=50 seeded realizations per case, true g=(+0.03,−0.02), Gaussian galaxy × Gaussian PSF, observations built by the module's own `make_ngmix_observation` path):
- *Direct fits, ivar weights*: mean g = (+0.0329, −0.0194) flat / (+0.0326, −0.0221) ramp — consistent with truth; pull std 1.08–1.12; **χ²/dof = 1.0027** under both flat and factor-8 ramp noise.
- *Binary weights under the ramp* (old MegaPipe-style): χ²/dof = **4.83**, paired shape-scatter ratio ivar/binary = 0.586 (asserted < 0.8).
- *Metacal end-to-end (noshear)*: ivar/binary scatter ratio 0.684 (asserted < 0.9) after `bb93c288`.

**The teeth** — injected wiring bugs, all caught: ×4 normalization → χ²/dof 4.0; 1/rms instead of 1/rms² → 13.2; inverted weights → 1.5×10⁵ (and scatter ratio > 1.6); transposed rms map → 13.9. One subtlety worth knowing: pull distributions alone can be laundered (ngmix rescales the parameter covariance by reduced χ²), so the absolute-normalization assertion rides on χ²/dof — the one statistic that can't hide a mis-scaled weight map.

**Honest boundaries**: the test enters at `make_ngmix_observation`, so vignetmaker's upstream extraction/orientation of the BACKGROUND_RMS vignettes is not exercised (the transposed-map probe shows χ² *would* catch a misalignment injected at the module boundary). Metacal pull std sits at ~1.4–1.5 even with perfect weights and flat noise — fixnoise/deconvolution correlated-noise bookkeeping inside ngmix, independent of the weights — so pull-calibration assertions live on the direct-fit layer.

11 tests, ~35 s; full suite 274 passed, no regressions. This file is also the natural home for the r50/T semantics regression checks going forward.

— Claude (Fable), on behalf of Cail

cailmdaley:
One more finding from tonight's star-validation investigation (prep for tomorrow's session with Fabian), orthogonal to the weight work but real: **the module's PSF-model fit is prior-dominated.** PSF observations are built with unit-flux stamps and no weight map, so the likelihood is much weaker than the prior — a PSF drawn with g1 = 0.025 fits to g1 = 9×10⁻⁵ (HSM confirms the stamp itself carries g1 = 0.0241; removing the prior recovers g = (0.0241, −0.0144)). Consequences: the `g_psfo`/`g_psf` catalogue columns are biased round, and the deconvolution kernel is rounder than the true PSF — a PSF-leakage risk. Present on both the current tip and the pre-June branch state, so it is not from the recent fixes. Happy to open this as its own issue with the reproduction script (`/n17data/cdaley/unions/scratch-wf/star-experiment/` on candide).

Possibly also relevant to Monday's m-bias discussion: the 2023 star-test failure (#656, still open) links to esheldon/ngmix#72 — metacal m ≈ −0.2 when the jacobian departs from a simple pixel scale. The WCS-passing question and the m-bias question may be the same thread; we're checking the WCS-delivery hypothesis against Fabian's run artifacts tomorrow.

— Claude (Fable), on behalf of Cail

martinkilbinger:
> One more finding from tonight's star-validation investigation (prep for tomorrow's session with Fabian), orthogonal to the weight work but real: **the module's PSF-model fit is prior-dominated.** PSF observations are built with unit-flux stamps and no weight map, so the likelihood is much weaker than the prior — a PSF drawn with g1 = 0.025 fits to g1 = 9×10⁻⁵ (HSM confirms the stamp itself carries g1 = 0.0241; removing the prior recovers g = (0.0241, −0.0144)). Consequences: the `g_psfo`/`g_psf` catalogue columns are biased round, and the deconvolution kernel is rounder than the true PSF — a PSF-leakage risk. Present on both the current tip and the pre-June branch state, so it is not from the recent fixes. Happy to open this as its own issue with the reproduction script (`/n17data/cdaley/unions/scratch-wf/star-experiment/` on candide).
> 
> Possibly also relevant to Monday's m-bias discussion: the 2023 star-test failure (#656, still open) links to [esheldon/ngmix#72](https://github.com/esheldon/ngmix/issues/72) — metacal m ≈ −0.2 when the jacobian departs from a simple pixel scale. The WCS-passing question and the m-bias question may be the same thread; we're checking the WCS-delivery hypothesis against Fabian's run artifacts tomorrow.
> 
> — Claude (Fable), on behalf of Cail

This could be a feature - with `fitgauss` an isotropic function is fitted (@fabianhervaspeters). To check.

aguinot:
Hello,
I just happened to look quickly over the changes in `ngmix.py` and I found a potential issue in the changes.
The `Tpsf` and `g1/2_psf` has been replaced by the `psfo` quantities (`o` stand for "original"). This is wrong and it is probably a lot better to keep both.
I would agree that `g1/2_psf` do not contain a lot of information as it is measured on the reconvolved PSF which in this case is an isotropic gaussian but it would be good to have as a sanitycheck. Also, the reconvolve PSF does not have to be isotropic so it could be useful to have if you plan on changing later.
Now `Tpsf` is the most problematic one. The reconvolved PSF is larger than the orignal one and you want to use that for the selection `Tgal/Tpsf` not the original PSF size. If you use `Tpsf_o` instead you will include smaller galaxies that would (should) be cuted off.
On a separate note, it is very dangerous to have a quantity that is not what it is suppose to be. Given that `g1_psf` is suppose to be the g1 of the PSF measured on the metacalibrated image it is expected to be 0. If you replace it by `g1_psf_o` then it will be =/= 0 and that could be very confusing.

I also saw a discussion on the prior for the PSF fit. I do agree that psf and galaxies should use different prior but it has never been an issue and I don't really see why it is an issue now. The size and flux are uniform over a range that cover the expected value for the PSF. The centroid should be fine too. So only the shape could be bias but it "roughly" a gaussian centered around 0 with an over estimated sigma for the PSF, which shouldn't be a big problem. 
I hope this can help!

cailmdaley:
**CI is green — we believe this PR is ready to merge.** ✅

`fast-tests`, `build-test-publish`, and the docs deploy all pass on `e57544be`.

A final pass since the last review round:

**Cleanups & review fixes**
- Renamed the `moments_fail` column → `mcal_types_fail` / `NGMIX_MCAL_TYPES_FAIL` through `ngmix` + `make_cat`. The v1 name counted `get_guess` (moments-guess) failures that no longer exist in the v2 fitter; it now counts the metacal types with nonzero fit flags, so the name now matches the meaning.
- A `/code-review` pass surfaced and fixed: a latent `NameError` in `make_cat._save_ngmix_data(moments=True)` (two arrays were read only in the non-moments branch but used unconditionally); a leftover debug `print` loop in `make_cat` (plus two stray prints in `ngmix` folded into a `ValueError`, and a missing f-string prefix); clear `ValueError`s in the `mccd_interp` / `read_ext_sexcat` / `psfex_interp` / `vignetmaker` runners when a MULTI-EPOCH/post-process config omits the positional WCS / exp-numbers input (previously a cryptic `IndexError` — no behaviour change for valid configs); and a divergence guard tying `compile_results`' metacal-type list to `METACAL_TYPES`.

**CI fix (`fast-tests` was red, unrelated to the science changes)**
- `pyproject.toml` had no `[build-system]` table, so uv resolved shapepipe as a *virtual* project (`uv.lock` recorded `source = { virtual = "." }`) and `uv sync` installed only the dependencies — every test module failed collection with `No module named shapepipe`. Declared the setuptools backend (lockfile now `editable`).
- `test_mpirun_launch_imports_mpi` now skips when the MPI runtime library is absent (the bare fast-tests runner; the dev-image CI still exercises the real MPI path), matching the workflow's self-skip pattern for system-dependent tests. `run.py`'s launcher gate is unchanged and correct.

**Deferred, intentionally out of scope for this PR**
- The PSF-fit prior-domination and `g_psfo`/`Tpsf` provenance issue is tracked separately in #749 (with a reproduction). Metacal *shear* is unaffected (ngmix deconvolves by the actual PSF image); only the PSF catalogue columns are, and those aren't currently ingested downstream.
- Non-blocking follow-ups noted during review, not done here: the `make_post_process` TILE_ID metadata-namespace special-case, a couple of per-object efficiency hot-spots in `ngmix.py` (per-epoch FSCALE header re-parse, per-object `make_runners`), and some numpydoc/PEP8 doc gaps on the new `ngmix` helpers.

— Claude, on behalf of Cail

---
_Generated by [Claude Code](https://claude.ai/code)_

aguinot:
Hello again Claude/Cail,

I am sorry to insist but the changes to `Tpsf` are wrong and not necessary. I totally understand that you want to keep the psf prior in a separate issue but [this change](https://github.com/CosmoStat/shapepipe/blob/e57544be49684a0a21aa98854ceec6053478914e/src/shapepipe/modules/ngmix_package/ngmix.py#L456) is introducing a bug and make it "impossible" to properly compute the selection response. I really don't understand why you are introducing it.
I will also answer to #749 with an example but as I suspected, the prior is not an issue.

cailmdaley:
**Update — consolidated to a single in-image CI path.**

Following a quick discussion: rather than keep the separate `fast-tests` job (which ran a subset in a bare `uv` venv, with the binary/MPI tests only self-skipping there), I've **removed `fast-tests.yml`**. This realigns with the test-suite constitution — *"the image is the environment… there is no second test workflow"* — and eliminates the environment divergence. `deploy-image.yml`'s in-container `pytest -rX` is a strict superset of what fast-tests ran, so no coverage is lost; it remains the gate.

Kept from the CI work: the `[build-system]` declaration (correct independent of CI — it makes `uv sync` / `pip install -e .` install the package, per the documented dev loop), and the MPI-launcher test's skip-when-unavailable guard (harmless in-image, where MPI is present).

All checks green on `216f407a`; PR is `mergeable_state: clean`.

— Claude, on behalf of Cail

---
_Generated by [Claude Code](https://claude.ai/code)_

martinkilbinger:
@cailmdaley we should maybe propagate both original and reconvolved PSF quantities Tpsf and Tpsf_o, following @aguinot's comment.

cailmdaley:
Thank you @aguinot for insisting! Sorry neither I nor Claude saw your initial comment before posting that "we believe this is ready to merge" comment. Will fix and take a look at your comment on #749. cheers

cailmdaley:
Okay @aguinot I dug into the history of this and will try to express my understanding below for the sake of educating myself, tell me if this sounds right to you.

Before this branch was opened both the original and reconvolved PSFs were fit, with the original `psf_obs` fit quantities appropriately assigned to `*o` keys:
https://github.com/CosmoStat/shapepipe/blob/0cfcf4f8a1ed4fa21bfb3cd47308f3133540aab9/src/shapepipe/modules/ngmix_package/ngmix.py#L754 https://github.com/CosmoStat/shapepipe/blob/0cfcf4f8a1ed4fa21bfb3cd47308f3133540aab9/src/shapepipe/modules/ngmix_package/ngmix.py#L798 and the reconvolved `obs.psf` fit quantities appropriately assigned to keys with no `o`: https://github.com/CosmoStat/shapepipe/blob/0cfcf4f8a1ed4fa21bfb3cd47308f3133540aab9/src/shapepipe/modules/ngmix_package/ngmix.py#L888-L896 https://github.com/CosmoStat/shapepipe/blob/0cfcf4f8a1ed4fa21bfb3cd47308f3133540aab9/src/shapepipe/modules/ngmix_package/ngmix.py#L907-L908 (here eventually `psf_res_gT` and `tres` get merged into a single dictionary).

---

Then, in d01503d06 the switch to ngmix 2.4 happened and `make_galsim_fit` went away, replaced by a `MetacalBoostraper` **that only fits the reconvolved PSF:** https://github.com/CosmoStat/shapepipe/blob/60962fff390f1a5b1ee065e6b0e37b5fe992d0c8/src/shapepipe/modules/ngmix_package/ngmix.py#L1187-L1189 and in 60962fff, the original `*o` keys that were no longer being calculated got assigned the reconvolved quantities:
https://github.com/CosmoStat/shapepipe/blob/60962fff390f1a5b1ee065e6b0e37b5fe992d0c8/src/shapepipe/modules/ngmix_package/ngmix.py#L688 Thus, both the `*psf` and `psfo` keys carried the same reconvolved PSF, and I agree that this is very dangerous. It already caused a lot of confusion for example below, where two `T_psfo` keys were defined:
https://github.com/CosmoStat/shapepipe/blob/60962fff390f1a5b1ee065e6b0e37b5fe992d0c8/src/shapepipe/modules/ngmix_package/ngmix.py#L410 https://github.com/CosmoStat/shapepipe/blob/60962fff390f1a5b1ee065e6b0e37b5fe992d0c8/src/shapepipe/modules/ngmix_package/ngmix.py#L422 which I then deduplicated in e57544b leaving the clobbered key as you noticed above. This didn't change any behavior, but it was a shortsighted change that didn't address the underlying issue and compounded confusion.

---
Anyway, it sounds like the solution is to correctly record both the original and reconvolved PSF as you suggest, and clean up these original-reconvolved clobbers. This would probably mean adding back an ngmix fit to the original PSF in `do_ngmix_metacal`, which would then be assigned to the *epoch-averaged* `NGMIX_ELL_PSFo` in the output catalog alongside the reconvolved `NGMIX_Tpsf`. 

However I'm a bit confused because there are also **per-epoch** moment fits to the **original** PSF e.g. `E1_PSF_HSM`, `SIGMA_PSF_HSM` which could in theory be inverse-variance averaged in the same way as the ngmix quantities. This creates some confusion about which product to use for what, e.g. the alpha-leakage calculation in `sp_validation` uses the `NGMIX_ELL_PSFo` whereas the rho/tau calculation uses `E1_PSF_HSM`. Is there any reason to use different fitting methods for the alpha vs. rho/tau stats?

Here's a concrete proposal, which may be stupid: make available inverse-variance averaged moment `*_HSM` fits for the original PSF, *and* `*_NGMIX` fits to both the original and reconvolved PSF, all with consistent naming and size definitions. This would give us 3/4 of the 2x2 matrix of original/reconvolved x HSM/NGMIX. Maybe this is overkill, please tell me what you think! Tagging @martinkilbinger @fabianhervaspeters @sachaguer since this touches lots of stuff.

—  Cail (written by himself!)

aguinot:
So I got confused by looking at the diff on github with [this line](https://github.com/CosmoStat/shapepipe/blob/216f407a0cdb9db2247ec94799f1b7fe56a015f3/src/shapepipe/modules/ngmix_package/ngmix.py#L456):
```python
output_dict[name]["Tpsf"].append(results[idx]["T_PSFo"])
```
Which is technically fine because before, [here](https://github.com/CosmoStat/shapepipe/blob/216f407a0cdb9db2247ec94799f1b7fe56a015f3/src/shapepipe/modules/ngmix_package/ngmix.py#L697-L698), you do:
```python
res['T_PSFo'] = psf_res['T_psf']
```
Which is extemely misleading to say the least...
So the situation is not as bad as I thought given that you store the most important quantity which is the reconvolved PSF. Now I would strongly encourage you to make the code more "human" readable and avoid this kind of misleading naming. And of course also store the fit of the original PSF.

Now, regaring your other questions. By computing the original PSF here we get the "coadded" PSF at the location of every detections in the catalogue, stars AND galaxies. The other PSF measurement you are refering too comes from the the star ONLY catalogue. So it cannot be used to compute the leakage at galaxy location, used for your object-wise PSF correction if I am not mistaken. In addition, the per-epoch with HSM is done in image coordinate while NGMix works in sky coordinates, which is a lot better (see my note at the end). If we wanted to "coadd" those PSF we would need to read the weight image at the star locations to get the weight values. Technically possible but a bookeeping nightmare and not useful. Probably require a specific module to do that, with matching, etc.
Summary:
- NGMix PSF: all detections "coadded" PSF in sky coordinate with the same fitter as for the shape measurement
- per-epoch PSF: stars only single epoch PSF in image coordinate with adaptive moments (close-ish to gaussian fitting)

I am sorry but don't understand your last proposal, I am not sure what usage we would have of this.

**NOTE about HSM:**
It is now possible to return the ellipticity in sky coordinate with HSM with the keyword: `use_sky_coords=True` which is `False` by default. I would strongly encourage you to change this aswell. Probably in a separate PR. (see [documentation](https://galsim-developers.github.io/GalSim/_build/html/hsm.html#galsim.hsm.FindAdaptiveMom))

cailmdaley:
Yes extremely confusing/misleading, it took me quite a while to trace this. I totally agree there should be a single source of truth for each quantity that is consistent with its column name and docstrings, although I'll note that all the non-human-readable elements were authored by humans :) 

I actually just discovered that the HSM moment fit is already calculated and saved at galaxy positions: https://github.com/CosmoStat/shapepipe/blob/216f407a0cdb9db2247ec94799f1b7fe56a015f3/src/shapepipe/modules/psfex_interp_package/psfex_interp.py#L312 https://github.com/CosmoStat/shapepipe/blob/216f407a0cdb9db2247ec94799f1b7fe56a015f3/src/shapepipe/modules/make_cat_package/make_cat.py#L631-L635 so if I'm understanding well, with the change to sky-coordinate HSM that you suggested one could get a per-object HSM fit by weighting the per-epoch fits in the same way that they're weighted for the ngmix fits without too much plumbing. But these quantities aren't used anywhere and in fact are discarded in sp_validation, so if it's not useful then let's not waste time on it!

So I will leave the HSM fits at galaxy positions alone (unless there are performance reasons to turn off the route altogether, since it doesn't seem to be used anywhere), clean up the original/convolved ngmix PSF keying, and open a separate PR for switching HSM fits to sky coordinates. Tell me if I'm still missing anything and I'll tag you when I have an implementation. Cheers

fabianhervaspeters:
_Hello, in the previous ngmix version the way we dealt with neighbors was to fill in a noise image were SExtractor gave the segmentation of a neighbor. https://github.com/CosmoStat/shapepipe/blob/fc656dae6709f15ded577ad0a48ee9818c2e3eca/src/shapepipe/modules/ngmix_package/ngmix.py#L1058
now we do not know if this strategy is perfect (for one thing it could vary a lot between exposures what the "neighbor" is) and we do have the undiagnosed B-modes when including FLAGS=2 galaxies, yet the idea was to propagate the same strategy in the new version for now, both because we lack a better alternative (since we are not going to implement uberseg) and because we fundamentally believe this should work. I discussed this with Eric Huff and he said that in Euclid the approach was to do a noise interpolation, so a very similar approach. The function get_noise is in ngmix.py but never used. Please let me know if I missed it.  Maybe we could build a mask in sky coordinates to make sure all exposures have the same area masked and then apply a 1 or 2 pixel broadening? Since discarding FLAGS=2 objects lead to a strong selection not-metaclibratable selection bias I think this deserves some thought._

EDIT: Sorry about that, I think this is okay for now, it actually appears as:

if (~mask).any():
        gal_masked[~mask] = noise_img_gal[~mask]

[line](https://github.com/CosmoStat/shapepipe/blob/0ae5b14927734ac060f77fa144bcda61dca354c7/src/shapepipe/modules/ngmix_package/ngmix.py#L1018-L1019)

aguinot:
@cailmdaley propagating to the coadd the single epoch HSM quantites will be a lot more difficult than you think and I don't see a good reason to do so.

aguinot:
@fabianhervaspeters I don't know why the `get_noise` function is not used anymore. It should give potantially more accurate results than `sigma_mad` used [here](https://github.com/CosmoStat/shapepipe/blob/0ae5b14927734ac060f77fa144bcda61dca354c7/src/shapepipe/modules/ngmix_package/ngmix.py#L990). The idea of `get_noise` is to mask the central object to estimate the noise. The stamps are only 51x51 pixels so in the case where there is a large object I wouldn't be surprise that sigma_mad is biased. But it will probably not make a big difference.
Regarding your idea of having a common mask between exposures, this is already the case. The segmentation map used is the one from the coadd. I wanted to make it slightly larger but given that the SNR is higher on the coadd I thought it would already enlarge the mask enough (maybe not..).

All that being said, I don't think we should mask and replace by noise neighboring objects on the image. A better startegy would be:
- replace by noise masked pixels (cosmic rays) as they will create a discontinuity when going to fourier space (metacal process). Replacing by noise is probably just as good as interpolation but likely faster (is the noise estimation is not to bad). AND set weights to 0.
- Leave the neighboring objects on the image as they are not an issue for metacal.
- Apply 0 weight for the neighbor and do uberseg.
- Probably do a symmetrization of the weights

For uberseg, it is really not difficult to implement. Here is a piece of code I made for another project to branch to the implementation of Erin Sheldon used in DES:
```python
import numpy as np

from meds import _uberseg

def fast_uberseg(
    seg,
    weight,
    object_number,
):
    obj_inds = np.where(seg != 0)

    Nx, Ny = seg.shape
    Ninds = len(obj_inds[0])
    seg = seg.astype(np.int32)
    weight = weight.astype(np.float32, copy=False)
    obj_inds_x = obj_inds[0].astype(np.int32, copy=False)
    obj_inds_y = obj_inds[1].astype(np.int32, copy=False)
    _uberseg.uberseg_tree(
        seg, weight, Nx, Ny, object_number, obj_inds_x, obj_inds_y, Ninds
    )

    return weight
```
For the symmetrization, it is implemented [here](https://github.com/esheldon/metadetect/blob/c92e11b8afe471d36c50794b137e76f6310995b5/metadetect/fitting.py#L1415).
I hope this can help.

(This is obvious but everything above need testing)

cailmdaley:
I just did one more review and cleaned up some leftover e-type HSM consumers that were missed in #761. Also found a preexisting issue #785 that needn't block this PR. @martinkilbinger I think this is now good to go!

martinkilbinger (approved):

PR #747, opened 2026-06-11
Gate mpi4py import on MPI launcher environment (fixes #744)

### LINKED ISSUE #744: [BUG] slurm on candide
I get an error when I run `shapepipe_run` on a compute node in the docker image. My commands:
```
srun --nodes=1 --ntasks-per-node=8 --time=24:00:00 --pty bash -i 
apptainer shell --bind /n17data,/n09data,/home /n17data/mkilbing/shapepipe_im_sims.sif 
shapepipe_run -h
```

The error message is:
```[n08:752859] OPAL ERROR: Unreachable in file ext3x_client.c at line 112
--------------------------------------------------------------------------
The application appears to have been direct launched using "srun",
but OMPI was not built with SLURM's PMI support and therefore cannot
execute. There are several options for building PMI support under
SLURM, depending upon the SLURM version you are using:

  version 16.05 or later: you can use SLURM's PMIx support. This
  requires that you configure and build SLURM --with-pmix.

  Versions earlier than 16.05: you must use either SLURM's PMI-1 or
  PMI-2 support. SLURM builds PMI-1 by default, or you can manually
  install PMI-2. You must then build Open MPI using --with-pmi pointing
  to the SLURM PMI library location.

Please configure as appropriate and try again.
--------------------------------------------------------------------------
*** An error occurred in MPI_Init_thread
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
[n08:752859] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!```

cailmdaley:
Fixes #744.

## The mechanism

`src/shapepipe/run.py` does `from mpi4py import MPI` at module import, which initializes MPI immediately — even for `shapepipe_run -h`. Inside an `srun`-launched shell, Open MPI sees the SLURM step environment, decides it was direct-launched by srun, and looks for a PMI server that srun never started (the container's OMPI is not built with SLURM PMI support). MPI_Init aborts the whole process before the pipeline ever runs.

Empirically bisected on candide: unsetting `SLURM_STEP_ID` alone is enough to make the unpatched code work — that's the variable OMPI keys its srun-detection on. `mpirun` works because it brings its own PMIx server.

## The fix

Only import (hence initialize) mpi4py when a launcher environment is actually present:

- `OMPI_COMM_WORLD_SIZE` — set by `mpirun`/`orterun`
- `PMI_RANK` — set by `srun --mpi=pmi2`
- `PMIX_RANK` — set by `srun --mpi=pmix`

A bare `shapepipe_run` (login node, compute-node shell, container) never touches MPI and runs SMP exactly as before; MPI launches are unchanged.

## Verification (on candide, inside the `ngmix_v2.0` container under a real `srun` allocation)

| scenario | unpatched | patched |
|---|---|---|
| bare `shapepipe_run -h` under srun shell | OPAL abort, exit 1 | **exit 0** |
| `mpirun -n 1 shapepipe_run -h` (the workaround from the issue) | exit 0 | exit 0 |
| `mpirun -n 2` → `import_mpi`, ranks | — | `True`, ranks 0/1 |
| bare on login node | exit 0 (singleton init) | exit 0 (no MPI touched) |

Plus a subprocess-based regression test (`test_run.py`) that scrubs/sets the launcher env and asserts the gate in both directions.

**Behavior note:** a config with `MODE = MPI` launched *without* any MPI launcher now falls back to SMP instead of running single-rank MPI. Arguably the saner behavior, but flagging it.

— Claude on behalf of Cail.

### DIFF
--- a/src/shapepipe/run.py
+++ b/src/shapepipe/run.py
@@ -6,6 +6,7 @@ This module sets up a given run of the shape measurement pipeline.
 
 """
 
+import os
 import sys
 from datetime import datetime
 from importlib.metadata import requires
@@ -22,12 +23,27 @@ from shapepipe.pipeline.file_handler import FileHandler
 from shapepipe.pipeline.job_handler import JobHandler
 from shapepipe.pipeline.mpi_run import split_mpi_jobs, submit_mpi_jobs
 
-try:
-    from mpi4py import MPI
-except ImportError:  # pragma: no cover
-    import_mpi = False
+# Importing mpi4py initializes MPI immediately, which aborts the whole
+# process when no MPI launcher is available — e.g. inside an
+# ``srun``-launched shell on a SLURM cluster, where Open MPI detects the
+# SLURM step environment, expects a PMI server that srun never started,
+# and calls MPI_Abort before even ``shapepipe_run -h`` can print (#744).
+# Only import (and hence initialize) MPI when a launcher environment is
+# actually present: ``mpirun``/``orterun`` set OMPI_COMM_WORLD_SIZE,
+# ``srun --mpi=pmi2`` sets PMI_RANK and ``srun --mpi=pmix`` sets
+# PMIX_RANK. A bare ``shapepipe_run`` (login node, compute-node shell,
+# container) runs in SMP mode without ever touching MPI.
+_MPI_LAUNCHER_VARS = ("OMPI_COMM_WORLD_SIZE", "PMI_RANK", "PMIX_RANK")
+
+if any(var in os.environ for var in _MPI_LAUNCHER_VARS):
+    try:
+        from mpi4py import MPI
+    except ImportError:  # pragma: no cover
+        import_mpi = False
+    else:
+        import_mpi = True
 else:
-    import_mpi = True
+    import_mpi = False
 
 
 class ShapePipe:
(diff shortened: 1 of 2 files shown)


cailmdaley:
An independent fresh-context review of the diff came back clean (no NameError path when `import_mpi` is False; no other unconditional mpi4py imports in the tree; all of this repo's launch paths — `candide_mpi.sh`, `cc_mpi.sh`, the documented `mpiexec` invocations — open the gate). Two of its nits are folded into `dab2d5c1`: the `module_dep` else-branch self-appended the whole list (pre-existing, but the gate made it the common path), and the regression test now surfaces subprocess stderr on failure.

One edge worth knowing, not a regression: `srun -n4 shapepipe_run` *without* a PMI-enabled MPI used to abort loudly; with the gate each rank now runs an independent SMP pipeline (N duplicate runs). A `SLURM_NTASKS > 1`-but-no-launcher warning would be a cheap follow-up if that ever bites.

— Claude on behalf of Cail.

cailmdaley:
Folded into `ngmix_v2.0` (6b30fd41, same commits — true merge), so the #744 fix ships with the ngmix v2.0 line regardless of review order here. If #741 lands first, GitHub will mark this PR merged automatically; reviewing either place covers both.

— Claude on behalf of Cail.

PR #748, opened 2026-06-11
Replace -e/--exclusive flag with NUMBER_LIST (#746)

### LINKED ISSUE #746: [HELP]
Can the exclusive input ID (-e flag) be replaced by the existing NUMBER_LIST?

cailmdaley:
Implements #746: the `-e`/`--exclusive` input-ID flag is retired; `NUMBER_LIST` is now the one mechanism for restricting a run to given IDs. Targets `ngmix_v2.0` so it rides #741 to develop (the v2.0 canfar scripts diverged from develop, so this is the conflict-free base).

**Pipeline**
- `-e`/`--exclusive` removed from `args.py`, with its plumbing through `run.py`, `FileHandler`, and `JobHandler` (where it was stored but never used)
- `NUMBER_LIST` entries are now validated against the input file numbers actually found on disk — preserving `-e`'s early failure on a typo'd ID: the run aborts at start-up with a clear message instead of when a module first tries to open non-existent files
- unit tests: subset passes, typo raises, scan path without `NUMBER_LIST` unchanged

**Canfar chain** — script-level `-e` options are **unchanged** (one ID per headless job remains the orchestration interface); what changes is how they realize it:
- `job_sp_canfar.bash`, `job_sp_canfar_v2.0.bash`, `init_run_exclusive_canfar.sh` write `NUMBER_LIST` into a per-job config copy (`set_config_number_list`: insert-or-replace under `[FILE]`; ID transformed to numbering-scheme form, leading dash + dots→dashes — exactly the transform `-e` did internally)
- side benefit: the processed ID now lands in the config that shapepipe copies into the run's log directory — better provenance than a CLI flag that vanishes

**Verified**: full test suite (61 passed) in the ngmix v2.0 container against this branch; `set_config_number_list` smoke-tested against `config_tile_Uz.ini` (insert), an already-set config (replace), and a `[FILE]`-followed-by-comment edge case.

Closes #746.

### DIFF
--- a/docs/source/configuration.md
+++ b/docs/source/configuration.md
@@ -84,7 +84,10 @@ The following options can be added to the `[FILE]` section of the config file
   (*e.g.* `.`, `-`, `:`, *etc.*). *optional*ly a regular expression can also be
   passed if it is preceded by `RE:` (*e.g.* `RE:-\d{9}`).
 - `NUMBER_LIST` : (`str` or `list`, *optional*) A list of number strings
-  matching the numbering scheme or a file name.
+  matching the numbering scheme or a file name. Restricts the run to these
+  numbers; every entry must match an input file found on disk, otherwise the
+  run fails at start-up. This is also how a single image is processed per
+  job (formerly the `-e`/`--exclusive` command-line flag).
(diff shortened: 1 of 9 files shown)


martinkilbinger:
Nice to see perl usage! :-)

Issue #749, opened 2026-06-12
[BUG] ngmix module passes the galaxy shape prior to the PSF fit, which is prior-dominated


cailmdaley:
This writes up the "PSF fit is prior-dominated" finding from PR #741 ([comment](https://github.com/CosmoStat/shapepipe/pull/741), 06-11) as its own thread, with the reproduction code @fabianhervaspeters asked for — and resolves the prior-vs-`fitgauss` question, including @fabianhervaspeters's full-pipeline result that removing the prior changed nothing. One claim from the original comment is also corrected below.

## What the code does

Two distinct PSF "fits" exist in the module, and the discussion conflated them:

1. **The PSF-model fit** — `make_runners` builds `psf_fitter = ngmix.fitting.Fitter(model='gauss', prior=prior)` ([`ngmix.py:1138`](https://github.com/CosmoStat/shapepipe/blob/ngmix_v2.0/src/shapepipe/modules/ngmix_package/ngmix.py)) where `prior` is the **galaxy** joint prior from `get_prior`, containing `GPriorBA(sigma=0.4)` (`ngmix.py:79`). The same construction exists pre-June (`bd60dc8e`, line 1061). The PSF `Observation` is built with **no weight map** (`ngmix.py:1029`; ngmix defaults to unit weights) on a unit-flux stamp.
2. **The metacal reconvolution kernel** — `metacal_pars = {'psf': 'fitgauss'}` (`ngmix.py:1193`). ngmix's `MetacalFitGaussPSF` does a *prior-free* admom fit and then builds a **round** `galsim.Gaussian`, dilated. Isotropic **by construction — and by design**; that part is a feature.

## What we measured

Setup (deterministic, N=30, seeds fixed; container `shapepipe_ngmix_v2.0-dev.sif`, ngmix 2.4.0): CFIS-like Gaussian PSF, fwhm 0.65″, sheared to g=(0.025, −0.015), drawn at 0.187″/px on a 48×48 unit-flux stamp. Pixel-window dilution makes the stamp's actual ellipticity g=(0.02406, −0.01444) — confirmed independently by HSM `FindAdaptiveMom`.

| Fit of the **original** PSF stamp | g1 | g2 | T |
|---|---|---|---|
| stamp truth (HSM) | 0.02406 | −0.01444 | — |
[...]

Issue #201, opened 2020-05-18
fixed bug in file handler - [merged]


sfarrens:
In GitLab by @sfarrens on Jan 22, 2019, 13:57

_Merges fildhd_bug_sf -> master_

- Fixed file matching bug in file handler

PR #761, opened 2026-06-20
ngmix v2.0: one column grammar across estimators (PSF source-of-truth, single size, E↔G)

### LINKED ISSUE #777: ngmix v2.0: one column grammar across estimators (PSF source-of-truth, single size, E↔G)
**Sub-issue of the ngmix v2.0 epic (#762).** Turns the PSF-column tangle that surfaced in #741 into **one column grammar for the whole catalogue**: every shape estimator the pipeline writes — ngmix, HSM moments, galsim — names its outputs `ESTIMATOR_COMPONENT_OBJECT[_metacaltype]` (uppercase), with a single size definition (`T = 2σ²`, sourced once from `cs_util.size`), ellipticity split into named scalar components with the e-type/g-type distinction made correct everywhere, `GAL` dropped, and `spread_model` removed.

This is the card for the work; the implementation rides the Development link below.

**Implemented by:** PR #761 (branch `refactor/psf-column-grammar`, base `ngmix_v2.0`) — mergeable, CI green. Depends on `cs_util.shape` (e↔g converters, #768 / cs_util#73) and `cs_util.size`.

Closes #761 is not used (the PR targets `ngmix_v2.0`, not the default branch); the Development link binds them on the board, and the chain auto-closes this issue when it reaches `develop`.

— Claude on behalf of Cail

### LINKED ISSUE #783: make_cat: remove the dead galsim shape-serialization path
Carved out of #761 per Axel's 2026-07-02 review ("don't change too many things at once").

`make_cat` still carries a galsim shape-serialization path — `_save_galsim_shapes()`, `GALSIM` as an allowed `SHAPE_MEASUREMENT_TYPE`, and a `test_galsim_grammar_properties.py` — but **no runner produces galsim shapes** (there is no `galsim_shapes_runner`), and the dual-shape (`len==5`) input branch in `make_cat_runner` never fires. ngmix is the pipeline's only shape estimator. The path is dead code.

This removes it:
- delete `_save_galsim_shapes()` from `make_cat.py` and drop `GALSIM` from `SHAPE_MEASUREMENT_TYPE` (make_cat_runner validates against `ngmix` alone, raising otherwise — the knob is kept as the extension point for a future estimator family)
- drop `test_galsim_grammar_properties.py` and the galsim case from the frozen column-grammar examples
- lose the galsim references in the docstring and the `cat_matched.param` comment

The removal is good work and stands on its own; it was pulled out of #761 only so that PR stays scoped to the reconvolved-PSF column grammar.

**Not blocking the ngmix v2.0 merge** — the galsim path is dormant either way, so this is cleanup, not a v2.0 correctness fix. Filed under the longer-term-improvements epic (#773).

Implementing PR is stacked on #761 (`refactor/psf-column-grammar`) because the removal deletes a test that #761 introduces; it retargets to `ngmix_v2.0` once #761 merges.

— Claude on behalf of Cail

cailmdaley:
ngmix: single source of truth for original vs reconvolved PSF columns

Follows the #741 review discussion of the ngmix PSF columns. The output conflated the original and reconvolved PSF — both `*_psf` and `*_psfo` carried the reconvolved fit, so `g1_psf` (expected ~0 on the metacal image) silently held the original-PSF value, and `Tpsf` did not mean what its name says. This PR makes each PSF quantity a single source of truth whose name matches its meaning, and unifies the catalogue column grammar across the ngmix and HSM star/PSF columns.

## Where to focus review
The science lives entirely in **`ngmix.py`** — specifically the two `average_*_psf` weighting functions (how each PSF family is combined across epochs) and the `do_ngmix_metacal` wiring (the metacal call + the new original-PSF pre-fit and its RNG handling). `make_cat.py`, `psfex_interp.py` and the `.param` files are mechanical serialization of the new column names; the rest is test coverage.

## Changes
- Restore an explicit fit to the **original** PSF and store **both** the original and reconvolved fits, split by name.
- Rename the ngmix columns to a consistent `ESTIMATOR_COMPONENT_OBJECT` grammar; update the shipped `.param` files.
- Unify the **HSM star/PSF column grammar** on the same convention (`E1/E2` for e-type, size stored as `T` via `cs_util.size`), including the fix to store the true e-type distortion in the `E*_HSM` columns. This is the "one grammar across estimators" that this PR exists for (issue #777, `E↔G` facet), and was reviewed with Sacha.
- Drop the dead `r50` columns (derivable from `T` via `cs_util.size`).
- Route the bare size conversions through `cs_util.size`.
- Property + unit tests for the new grammar and both PSF families.

## Scope (updated after Axel's 2026-07-02 review)
This PR is now held to the reconvolved-PSF / column-grammar purpose. Two things it previously carried have been pulled back out:
- **spread_model is restored** — it is no longer touched by this PR. Axel's point was that spread_model must not be *cut on* for shear selection, not that it be deleted; whether it should be removed at all is a separate question, handled outside this PR.
- **The galsim dead-path removal moves to #784** (Closes #783), stacked on this branch and retargeting to `ngmix_v2.0` once this merges.

## Notes for review
- The original-PSF pre-fit is seeded from a **snapshot of the metacal RNG**, so it does not advance the RNG metacal consumes — the galaxy/shear results are **bit-identical to the pre-PR base**; only the new `*_PSF_ORIG` columns are added. (Rationale in the `do_ngmix_metacal` comment.)
- Base: `ngmix_v2.0` — lands in #741 before the v2.0 merge.
- Partially addresses #749 (the column provenance / naming facet).
- This renames/splits catalogue columns with a real downstream surface in `sp_validation` — tracked in a separate draft PR there (#201).

— Claude on behalf of Cail

Closes #777


cailmdaley:
Hello @aguinot @martinkilbinger, this is my first attempt at a fix to all the PSF column issues that cropped up in #741. I decided to be very ambitious and define a common grammar for all column names (not just the PSF columns) to reduce ambiguity; this is quite far reaching since it changes everything that consumes the catalog, and will require a second PR in sp_validation to propagate these changes. However I think it will be worth it and is largely mechanical to propagate. If you think this is too aggressive please let me know and we can consider descoping to only touching the PSF columns. The only real science changes are the new ngmix fit to the original PSF (and the averaging over epochs) in ngmix.py.

Cheers, Cail

aguinot:
Hello,
I would remove `GAL` from the ngmix quantities as the objects are not necessarily glaxies and the name let's suppose that a selection has been done and that you have a sample of galaxies only safe to use, while it is not the case. And it also makes the names longer.
You can also keep the spread model up-to-date but, it should not be used to apply a cut as it will create selection effects not taken into account by the response matrix.
I would also use a single size definition everywhere to avoid confusion. If you decide to use the trace `T` (which I strongly recommend!) only use that and remove columns such as `PSF_FWHM` (or converrt them to `T`), it can be derived from `T`.
Finally, I don't know why I used full capital names in the output catalogue but it might not be a good idea. Feel free to use lower case names.

sachaguer:
Some notes on the column labelling for the star catalogue. I discussed with @fabianhervaspeters that the HSM shapes are e-types ellipticities, whereas ngmix returns g-type ellipticities. The new labelling shows that, for ngmix, we should ensure the naming is consistent for the HSM quantities and potentially add a function to easily convert between them.

cailmdaley:
This standardizes the output catalogue onto one column grammar. Every shape estimator the pipeline writes — ngmix, galsim, HSM moments — now names its columns the same way: `ESTIMATOR_COMPONENT_OBJECT[_metacaltype]` (uppercase), with the galaxy as the implicit default object (no token), ellipticity split into named scalar components (`G1/G2` for g-type, `E1/E2` for e-type), and a single stored size `T = 2σ²` from `cs_util.size`. Any column then parses unambiguously into estimator · quantity · object · metacal shear.

**What changed:**
- `98e822ad` — HSM e/g value fix (below)
- `4b17f7ec` — HSM star/PSF grammar + size stored as `T`
- `85a18ab5` — remove the unused `spread_model` module
- `4f0756d0` — make_cat grammar for ngmix, galsim & the HSM epoch-sink columns (drop `GAL`, split packed ellipticity, `SIGMA`→`T`, fix a latent `n_obj` NameError)
- `f9a8f050` — track `cs_util @ develop` for the `cs_util.size` helpers

**Representative renames:**
- `NGMIX_G1_GAL_NOSHEAR` → `NGMIX_G1_NOSHEAR` (and the rest of the ngmix galaxy block)
- `E1_PSF_HSM` → `HSM_E1_PSF`, `SIGMA_PSF_HSM` → `HSM_T_PSF`, `FLAG_PSF_HSM` → `HSM_FLAG_PSF` (+ STAR variants)
- `GALSIM_GAL_ELL_<ext>` (packed length-2) → scalar `GALSIM_G1_<ext>` / `GALSIM_G2_<ext>`; `GALSIM_GAL_SIGMA_<ext>` → `GALSIM_T_<ext>`

**One value change (`98e822ad`).** The HSM star/PSF columns are named `E*` (e-type distortion) but were storing galsim's reduced shear `.g1/.g2` (g-type) — name and value disagreed. They now store `.e1/.e2`, so each column holds the quantity its name says. This is the only change that moves numbers downstream (the ρ/τ star–PSF residual statistics).

— Claude on behalf of Cail

sachaguer:
I don't understand what Claude says here:

> One value change (98e822ad). The HSM star/PSF columns are named E* (e-type distortion) but were storing galsim's reduced shear .g1/.g2 (g-type) — name and value disagreed. They now store .e1/.e2, so each column holds the quantity its name says. This is the only change that moves numbers downstream (the ρ/τ star–PSF residual statistics).

I think the HSM columns stored e-types ellipticities that were not converted to g-type. My only point was about having consistent labeling across `ngmix` shapes and `HSM` shapes. Can we confirm this?

**Edit**: I checked the commit. It appears that what we said with Fabian was incorrect and that we saved g-type ellipticities. This is actually fine and what we need for rho/tau-statistics. We could go back to saving g-type and naming the columns appropriately. Would we need to save the e-type as well? Probably not if we have built-in functions to easily make the conversion.

cailmdaley:
Thanks @sachaguer, this looks great. The argument for recording `e` for HSM would that HSM natively measures e-type quantities, and since `e <-> g` is a nonlinear transformation, if your HSM `e`-type errors are Gaussian, the same wouldn't necessarily be true for your `g`-type errors. Jensen's inequality would also say their population means would be different. However for PSFs/stars the ellipticity is so small, we have to use `g`-type for leakage anyway, and as you say we have a lossless conversion function so let's leave it as is.

I think the last thing on this PR is for a shapepipe expert to take a look at the changes to ngmix.py since I added new functionality to average multi-epoch PSF measurements. @martinkilbinger @aguinot @fabianhervaspeters, could one of you please review that when you have a chance? Once this is merged then we should be able to finally merge #741!

sachaguer:
Agreed on the noise properties. I think this product is mainly used for leakage estimation. If any other use I am not aware of would be simplified by saving both quantities, I don't mind saving both.

aguinot:
Hey,
Just a small comment, I think it would be a lot easier to keep the ellipticity in the convention throughout the code, as much as possible. I would store the star and PSF as reduced shear g1/g2 instead of e1/e2. When you will want to compute cross-correlation between PSF and galaxies you will have to convert them back.
Most ngmix fitters will return g-type ellipticities but in any case once calibrated with the response you will always have a g-type.

cailmdaley:
Thanks Axel, I think your suggestion is what Sacha implemented.

cailmdaley:
**Follow-up: remove the dead galsim shape path**

#761 brought the `GALSIM_*` output columns under the v2 grammar — but those columns can't actually be produced. ShapePipe has no galsim shape-measurement runner; only `ngmix_runner` measures shapes and writes shape columns. The `GALSIM_*` family could only be filled by `make_cat`'s dormant `SHAPE_MEASUREMENT_TYPE=galsim` mode, which reads an external shape catalogue that nothing upstream generates — and every production config pins `ngmix`. The sp_validation side reached the same conclusion independently (its mirror galsim metacal consumer was removed as dead code, sp_validation #201). So rather than ship grammar-renamed columns no pipeline can exercise, this removes the galsim shape path entirely.

- `e036a95a` — delete `SaveCatalogue._save_galsim_shapes` and its `process()` "galsim" branch (modes are now `ngmix` / `psf`); drop the now-orphaned `cs_util.size` import; delete `test_galsim_grammar_properties.py`
- `df84064f` — `make_cat_runner` validates `SHAPE_MEASUREMENT_TYPE` against `ngmix` only; the shape-write loop drops the galsim-vs-ngmix catalogue selection; docstring / param comments lose their galsim references
- `1c69af55` — drop galsim from the frozen column-grammar test (regex branch + example columns); it now spans the two families ShapePipe actually produces, ngmix + HSM

Full suite green in-container (310 passed, 2 deselected). `galsim`-the-library uses — ngmix jacobians, `galsim.hsm` for HSM shapes, MCCD, PSFEx, the test image simulation — are untouched.

Considered but not done: `SHAPE_MEASUREMENT_TYPE` is now single-valued (`ngmix`). It could be dropped entirely and ngmix hardcoded, but that touches the collaboration's `make_cat` config API, so I kept it as a validated knob (the natural extension point for a future estimator, mirroring sp_validation's retained `prefix`). Happy to rip it out if you'd prefer.

— Claude on behalf of Cail

martinkilbinger (approved):

aguinot:
Hello,
I just have some generic comments:
- I think you should be carreful not to change too many things at once to keep potential bug trackable. This PR is about the reconvolved PSF and there are changes in the star PSF columns
- Most importantly, I said that the spread model should not be used to make galaxy selection for shear. But it doesn't mean that it should be deleted from ShapePipe. Is there a reason to do that? Is the code not working? I believe it could still be useful for otherthings. In any case it does not hurt to keep it in shapepipe. Also, this is not the right PR to remove it as it has nothing to do with the reconvolved PSF.

cailmdaley:
Thanks Axel. I split off the galsim cleanup to keep this PR focused, and renamed it to reflect the actual scope as tracked by issue #777. It would be very messy to separate the star and psf column renames into different PRs at this point. I also restored the spread model module and made it work with the new column names.

I think we can merge this PR once CI finishes.

PR #770, opened 2026-06-30
ngmix: config-selectable UberSeg neighbour masking (BLEND_HANDLING)

### LINKED ISSUE #776: UberSeg neighbour masking (Erin Sheldon) — config-selectable, validated against noise-fill
## What

Add a config-selectable **UberSeg** neighbour treatment to the ngmix module, alongside the current noise-fill, and decide how the segmentation map reaches the per-epoch stamps.

Today a neighbour sharing a galaxy's postage stamp is handled by **filling the neighbour's pixels with noise** (`prepare_ngmix_weights`): the per-epoch flag vignet marks them and they are replaced by a Gaussian realisation before the stamp reaches ngmix. The 2026-06-23 tomography call flagged this as suspect — including blended (`FLAGS=2`) objects produces **B-modes**, while discarding them is a strong, non-metacalibratable selection bias. UberSeg (Sheldon's MEDS/ngmix treatment) is the principled alternative: *mask* (weight → 0) every stamp pixel closer to a neighbour than to the central object, rather than synthesising noise over it.

## Implementation status

The option is implemented additively, default unchanged, in PR #770:

- `BLEND_HANDLING = {noisefill, uberseg}`, default `noisefill`, read like `CENTROID_SOURCE`. With the default the behaviour is **byte-for-byte identical to today** (asserted in tests).
- `uberseg_weight(weight, seg, object_number)` reimplements Sheldon's `meds._uberseg.uberseg_tree` nearest-segment Voronoi partition with a scipy `cKDTree`. We **reimplement rather than depend on `meds`**: the astronomy MEDS is `esheldon/meds` (not the PyPI `meds`, which is an unrelated package), and `import meds` drags the full `fitsio`/`esutil` stack for a single C tree. The surviving central core is a single connected, roughly circular region — emergent geometry of the partition, faithful to MEDS `get_uberseg` (no separate circular aperture).
- Plumbed runner → `Ngmix` → `do_ngmix_metacal` → `make_ngmix_observation` → `prepare_ngmix_weights`; `Postage_stamp` gains a per-epoch `segs` list.
- Unit tests: synthetic two-object stamp (mask geometry, single-connected core, matches brute-force nearest-segment), noisefill byte-identity, uberseg leaves the image untouched while hard-masking the weight.

## ⚠️ Open decision — the segmentation-map source

UberSeg needs the **segmentation map with object IDs** at each stamp, plus the central object's NUMBER. ShapePipe passes only a **binary** flag vignet to ngmix today, cut per-exposure (multi-epoch). The segmentation map with NUMBERs is the **coadd's** (the tile SExtractor detection) — but ngmix fits **per-epoch exposure stamps**, in different pixel frames. MEDS itself resolves this by *interpolating the coadd seg into each single-epoch frame* (`interpolate_coadd_seg`). So the source is genuinely forked:

1. **Extend `vignetmaker` to cut a coadd `SEGMENTATION` vignet** (the tile SExtractor already has the `CHECKIMAGE` machinery; it currently emits only `BACKGROUND`) and reproject it into each epoch frame via the per-epoch WCS — Sheldon-faithful, but needs nearest-neighbour resampling of an integer seg map.
2. **Approximate from the existing per-epoch binary flag vignet** — no reprojection, but the binary flag carries no per-object IDs, so the "closer to neighbour than to central" distance transform cannot be computed faithfully (it conflates neighbours with cosmic rays / bad pixels).

PR #770 implements everything **not** gated by this (the algorithm, the `BLEND_HANDLING` plumbing, the consumption seam, the unit tests); selecting `uberseg` without a seg map raises with a pointer here. This decision wants the shape-measurement experts (Axel Guinot, Martin Kilbinger, Fabian Hervas Peters): which seg source, and is per-epoch reprojection of the coadd seg the right path?

## Science note for the A/B validation

Following the explicit design intent (and Axel's "leave neighbour objects on the image" proposal), the `uberseg` path **zeroes the weight and leaves the image untouched** — it does *not* noise-fill the masked region. The open consideration: metacal shears the image in Fourier space using all pixels regardless of weight, so residual neighbour flux can leak through the shear even at weight 0. This is exactly what the noise-fill-vs-uberseg **A/B B-mode comparison** (on a `FLAGS=2`-inclusive sample, via the Paper II / sp_validation B-mode machinery) must measure — and what Axel's weight-symmetrisation follow-on (out of scope here) would address. Fabian validates the science before any production default flip.

## Scope fence

Add the option, its tests, the consumption seam; resolve the seg source; then validate. **Not** in scope: changing the noise-fill path, flipping the production default, the weight-symmetrisation / cosmic-ray-only-noise refinement, or the `get_noise`-vs-`sigma_mad` noise-estimate question (that belongs with #769).

Implemented (un-gated parts) by **PR #770** (`Closes #776`). Part of ShapePipe **longer-term improvements (#773)** — not blocking the ngmix v2.0 merge.

— Claude, on behalf of Cail

cailmdaley:
Closes #776.

## What

A config-selectable **UberSeg** neighbour treatment for the ngmix module, wired end-to-end and off by default:

- **`BLEND_HANDLING = {noisefill, uberseg}`** (default `noisefill` — the historical behaviour, byte-for-byte unchanged, asserted in tests). `uberseg` hard-masks (weight → 0) every stamp pixel closer to a neighbour's segmentation footprint than to the central object's, leaving the image untouched — Erin Sheldon's MEDS/ngmix treatment.
- **The segmentation input, per @aguinot's answer on #776**: the tile SExtractor run emits `SEGMENTATION` (one config line); a CLASSIC vignetmaker run cuts integer, non-interpolated seg stamps in the **coadd frame**; that single stamp is overlaid unchanged on every per-epoch stamp — **no reprojection** (UNIONS coadds carry no field rotation vs single exposures, so residual offsets are a few px at most).
- **`DILATE_NEIGHBOUR`** (int, default 1): binary dilation of the neighbour footprints to absorb those few-px offsets. Geometrically self-targeting: for well-separated objects the Voronoi bisector already claims the whole neighbour side and dilation adds nothing; it only bites for close blends — exactly the offset-sensitive regime.
- **`NGMIX_NEIGHBOUR_FLAG`** propagated to the final catalogue (does the seg stamp contain any non-central object?), enabling downstream systematics splits.

## Design notes

- `uberseg_weight` reimplements `meds._uberseg.uberseg_tree` (nearest-segment Voronoi partition) with a scipy `cKDTree`, **validated bit-for-bit against Sheldon's reference `get_uberseg`** (only measure-zero boundary ties differ). Reimplemented rather than depended on: `esheldon/meds` drags the `fitsio`/`esutil` stack for a single small algorithm.
- The central object is identified by its SExtractor `NUMBER` (seg labels *are* NUMBERs from the same run) — robust to registration offsets, which could otherwise let a neighbour's label sit on the centre pixel and invert the mask. The centre-pixel label is kept as a per-object diagnostic (warns on mismatch; the object is dropped if its footprint is absent).
- Seg catalogue row-alignment with the tile catalogue is asserted at load (fail fast on length or `NUMBER` mismatch).

## Why (fitting, not shearing)

Neighbour handling matters for the **model fit**: masking strictly along one side of a galaxy makes the weight anisotropic and lets the fit be "pulled" from the unmasked side. UberSeg reduces that pull by construction; 4-fold **weight symmetrisation** (@aguinot's least-biased endpoint) removes it entirely and is the declared follow-on PR — it composes with, and is orthogonal to, this one. The metacal *shearing* side (discontinuities, noise-fill of flagged pixels) is untouched here.

## Validation & default

Unit tests cover the mask geometry (vs brute force and vs Sheldon), byte-identity of the default, dilation additivity, the fail-fast guards, and the flag reaching the final catalogue (56 ngmix/make_cat tests; full module+unit sweep green). The production default stays `noisefill` until a real-data A/B (FLAGS=2-inclusive B-modes) says otherwise — that comparison is tracked separately and is not a merge condition for this dormant, default-unchanged capability.

— Claude, on behalf of Cail


aguinot:
Hello,
I just want to mentioned that I looked again at the DES Y6 config file and they do not symmetrize the `uberseg` weights. I am not sure I know why.. I am discussion with the people who did the analysis to understand this change.

aguinot:
Regarding the implementation, I see that you are doing a pure python with KDTree. What is the compute time? Is it trackable?

cailmdaley:
here's what Claude found, scipy cKDTree uses Cython and the inner loop is C++:

```
┌───────────────────────────────────┬────────────┬──────────────────┬───────┐
│              n_neigh              │ ours (μs)  │ Sheldon's C (μs) │ ratio │
├───────────────────────────────────┼────────────┼──────────────────┼───────┤
│ 0 (isolated, 45% of real objects) │ 26.6 ± 0.1 │ 39.7 ± 0.1       │ 0.67× │
├───────────────────────────────────┼────────────┼──────────────────┼───────┤
│ 1 (35% of real objects)           │ 990 ± 71   │ 1094 ± 44        │ 0.91× │
├───────────────────────────────────┼────────────┼──────────────────┼───────┤
│ 2 (14% of real objects)           │ 1059 ± 77  │ 1155 ± 69        │ 0.92× │
├───────────────────────────────────┼────────────┼──────────────────┼───────┤
│ 3 (4% of real objects)            │ 1168 ± 88  │ 1206 ± 46        │ 0.97× │
├───────────────────────────────────┼────────────┼──────────────────┼───────┤
│ 4 (rare)                          │ 1214 ± 52  │ 1226 ± 58        │ 0.99× │
└───────────────────────────────────┴────────────┴──────────────────┴───────┘
```
> We compared uberseg_weight()'s cKDTree implementation against your meds._uberseg.uberseg_tree reference at production stamp size (51×51px), averaging timing over 20 independent random neighbour geometries per neighbour count to get a stable estimate, and also summed the cost over every real object (30,435) in one actual UNIONS tile using real catalogue positions. The two are effectively tied — ours runs 5–10% faster in the realistic 0–2-neighbour regime that covers 94% of real objects (isolated objects, the 45% majority, are ~33% faster for us specifically), and the whole real tile's UberSeg pass costs under 20 seconds either way.

cailmdaley:
thank you for the comment and checking with DES folks, let me know if you have any more comments on this PR!

aguinot:
Interesting, I was not expecting scipy to be faster, good to know!

cailmdaley:
Merged into develop with origin/develop merged in first (through #843; all conflict resolutions additive). As validation, a noisefill control run of this exact merged tree on the image simulations reproduces the baseline catalog bit-for-bit — the new code is a verified no-op when `BLEND_HANDLING` is left at its default. The uberseg A/B characterization on the image sims is in progress under #813. Unit tests (20) and CI pass on the merged head.

— Claude (Fable), on behalf of Cail.

Issue #774, opened 2026-06-30
ngmix: restore (or justify dropping) the PSF noise-injection / PSF-observation weight


cailmdaley:
A small PSF noise-injection present in the previous (pre-v2.0) code is **absent in the v2.0 module** (Fabian's observation). The production PSF observation is built **weightless and noiseless** — `psf_obs = Observation(psf, jacobian=psf_jacob)` (`src/shapepipe/modules/ngmix_package/ngmix.py:1100`) — in contrast to the galaxy observation, which carries both `weight` and `noise` (`ngmix.py:1143–1148`). The older esheldon/aguinot pattern instead gave the PSF stamp a small injected noise and a matching finite weight (`psf_noise = 1.0e-6`, `psf_wt = 1.0/psf_noise**2`).

It survived only in dead example scripts, removed in `bd60dc8e`; the v2.0 module was a from-scratch rewrite that built the PSF observation weightless from the start (since `2b755516`), so there's no single "removal" commit — the exact prior version Fabian recalls is most likely the pre-v2.0 `aguinot/ngmix` 1.3.6 fork. **First step: pin which version + diff he saw, with Fabian.**

**Why it matters (and is pre-merge relevant):** a weightless PSF observation leaves ngmix assuming a default (effectively uniform) weight for the PSF stamp, so the noise the PSF fit assumes differs from the old `1/psf_noise²` budget. That feeds the metacal error bookkeeping and could change the star response R and the shape-measurement error model.

**Checkable on the twin:** re-introduce the PSF noise/weight as an additive knob, sweep it, and watch (a) the star response R and (b) χ²_red — does restoring it move either toward sanity, or was it correctly dropped?

Coupled to the reduced-χ² anomaly (#769, longer-term under #773): a weightless PSF observation is a leading candidate for the brightness-dependent error-model mis-specification that anomaly measures. Part of the ngmix v2.0 epic (#762).

— Claude on behalf of Cail

cailmdaley:
[...]

Issue #769, opened 2026-06-30
ngmix: the reduced-χ² anomaly on stars (error model)


cailmdaley:
On stars, reduced-χ² runs **too large at the bright end** (climbing to ~8 for the brightest) and **too small at the faint end** (~0.2, against an expected ~1). χ² naturally improves for faint objects (noise dominates the denominator), but ~0.2 is well below 1 and points at a mis-estimated error model or an over-flexible fit. The faint-star ≈0.2 is the genuinely concerning piece: it implies the assumed variance is ~5× too large (σ inflated ~2.2×).

The sign **flips with brightness**, so a single noise-mis-scaling can't explain both tails — the answer likely names more than one effect:
- **bright-high** reads as the Gaussian model genuinely failing to represent a near-point source through an imperfect PSF model (model inadequacy at the deconvolution singularity — consistent with the star-response R-climb floor);
- **faint-≈0.2** reads as the *assumed errors being too large* — which ties to the weight map / noise estimate fed to ngmix and to the #749 prior/weight thread.

**Checkable:** χ²_red vs magnitude re-derived from a controlled rerun, split by branch; the noise/weight path the fit actually consumes audited end-to-end (`prepare_ngmix_weights` `ngmix.py:986`, `get_noise` `ngmix.py:941`); and, where fixable, the assumed errors corrected so χ²_red → ~1, or the residual named as an intrinsic model-inadequacy floor.

**χ² is not exported today** — the fit result dict surfaces only `s2n` (`ngmix.py:505–512`); χ² lives inside the ngmix LM result. A reduced-χ² column needs wiring out per object (the prototype `feat(ngmix): write reduced chi2 …`, commit `ba384428` on `investigate/star-response`, is the clean seed). That export is what unblocks the χ²-vs-magnitude panel the verification suite (#765) wants. Also fix the χ²-vs-mag plot's **"Galaxy Fit"** y-axis mislabel — it sits on a *star* fit.

**Longer-term** (not blocking the ngmix v2.0 merge — under #773). Coupled to the dropped PSF noise-injection (its own issue under #762), which is a leading candidate for the error-model mis-specification this anomaly measures. Related: #749, #765 (consumes the χ² export).

— Claude on behalf of Cail

PR #779, opened 2026-06-30
ngmix: give the PSF observation a finite weight map (#749)

### LINKED ISSUE #749: [BUG] ngmix module passes the galaxy shape prior to the PSF fit, which is prior-dominated

cailmdaley:
Closes #749 

## What

`make_ngmix_observation` built the PSF observation **weightless**
(`psf_obs = Observation(psf, jacobian=psf_jacob)`), while the galaxy
observation right below it carries a full weight map. With no weight, ngmix
defaults to unit weight on the unit-flux PSF stamp, so the galaxy
`GPriorBA(0.4)` prior handed to the PSF fitter swamps the (tiny) likelihood and
the recovered PSF **shape** collapses toward the prior (zero ellipticity).

This PR gives the PSF observation a finite flat weight map
`psf_wt = 1/PSF_NOISE**2` (`PSF_NOISE = 1e-5`, the esheldon/aguinot value
Axel's #749 reproduction used) — the PSF analogue of the galaxy weight built
just below. Weight only: PSFEx/MCCD models already carry a little noise, so no
explicit stamp-noise injection is added (Axel's guidance on #749). The galaxy
prior is left in place (option B from the thread) — once a real weight is
present the likelihood reasserts and the prior no longer bites.

Closes #749

## One root, two issues

The single weightless line was the root of both #749 (the galaxy shape prior
prior-dominates the PSF fit, collapsing the recovered ellipticity) and the now
-retired #774 (the small finite-weight noise budget of the pre-v2.0
esheldon/aguinot code was dropped in the rewrite). They are the same bug from
two angles; the finite `1/psf_noise**2` weight **is** the substance of the old
noise-injection, so adding the weight resolves both. #774 carried no separate
deliverable and has been retired as redundant.

## Evidence — end to end through `do_ngmix_metacal` / `average_original_psf`

A sheared-PSF stamp (`psf_shear = (0.05, -0.03)`, HSM moment truth
`g = [+0.0477, -0.0286]`) pushed through the real production path; the
`*_psf_orig` column is the original-PSF fit:

| build | `g_psf_orig` | frac recovered |
|---|---|---|
| weightless (bug) | `[+0.0001, -0.0001]` | **0.00** |
| weighted (this PR) | `[+0.0477, -0.0286]` | **1.00** |

`g_psf_reconv ≈ 0` in both builds — the reconvolution kernel is round by
construction, independent of this weight (and is the column the pre-#761
`g1_psfo_ngmix` name was mis-sourcing).

- **Calibration is provably untouched.** Every metacal-type shear
  (`noshear/1p/1m/2p/2m`) is **bit-for-bit identical** (Δ = 0) between the
  unit-weight (pre-fix) and weighted builds, and the fit is flat across
  `PSF_NOISE ∈ {1e-4, 1e-5, 1e-6}`. Metacal fits this same `psf_obs` with a
  prior-free `AdmomFitter` (`MetacalFitGaussPSF._do_psf_fit`), which is
  insensitive to a flat weight's absolute scale — so the diagnostic weight
  never reaches calibration. It is the **galaxy prior** on the diagnostic fit
  (`average_original_psf`), not weight-independence, that makes `*_psf_orig`
  sensitive while the calibration is not.
- **Weight-only vs weight+noise (#774's question).** Injecting explicit
  PSF-stamp noise (σ = 1e-6, 1e-5) on top of the weight moves `g_psf`, `T_psf`,
  and `R` by ≲ 1e-6 — explicit injection is justifiably dropped.

The faint-end reduced-χ² / star-response coupling #774 also speculated about is
tracked separately under the χ²-anomaly investigation (#769), not here.

## Tests

`tests/module/test_ngmix_weight_validation.py` gains three guards: a structural
check that `psf_obs` carries the finite weight (never unit weight), the #749
end-to-end recovery contrast through `average_original_psf` (weighted recovers
to 1e-3, weightless collapses below 10%), and the bit-for-bit metacal-shear
invariance.

— Claude on behalf of Cail

### DIFF
--- a/src/shapepipe/modules/ngmix_package/ngmix.py
+++ b/src/shapepipe/modules/ngmix_package/ngmix.py
@@ -23,6 +23,15 @@ from shapepipe.pipeline import file_io
 
 METACAL_TYPES = ('noshear', '1p', '1m', '2p', '2m')
 
+# Noise budget for the PSF observation's flat weight map (psf_wt =
+# 1/PSF_NOISE**2). Mirrors the esheldon/aguinot pattern (sigma ~ 1e-5/1e-6);
+# 1e-5 is the value Axel Guinot's #749 reproduction used. The fit is driven
+# by the *relative* weighting of the likelihood vs the prior, so the exact
+# value is non-critical once it is finite (validated on the digital twin: the
+# recovered PSF shape/size are flat across 1e-4..1e-6). See
+# make_ngmix_observation.
+PSF_NOISE = 1e-5
+
 
 class MetacalResult(NamedTuple):
     """Return of :func:`do_ngmix_metacal`: the metacal fit plus both PSFs.
@@ -1091,7 +1100,19 @@ def make_ngmix_observation(
         col=(psf.shape[1] - 1) / 2,
         wcs=wcs,
     )
-    psf_obs = Observation(psf, jacobian=psf_jacob)
+    # A model fit always needs a weight: without one ngmix defaults to unit
+    # weights on a unit-flux PSF stamp, so the galaxy GPriorBA(0.4) prior
+    # handed to the diagnostic PSF fitter swamps the (tiny) likelihood and the
+    # recovered PSF shape collapses toward the prior (#749). PSFEx/MCCD models
+    # already carry a little noise, so a finite flat weight matching the
+    # esheldon/aguinot noise budget (psf_noise ~ 1e-5) suffices to restore the
+    # likelihood — no explicit stamp-noise injection needed (#774). This is the
+    # PSF analogue of the galaxy weight built just below; force float so an
+    # integer-typed stamp cannot truncate the weight (matching that build).
+    # Metacal's own PSF fit is prior-free (AdmomFitter) and so is insensitive
+    # to this flat weight's scale, leaving the calibration untouched.
+    psf_wt = np.full_like(psf, 1.0 / PSF_NOISE ** 2, dtype=float)
+    psf_obs = Observation(psf, weight=psf_wt, jacobian=psf_jacob)
 
     gal_masked, weight_map, noise_img = prepare_ngmix_weights(
         gal, weight, flag, rng, bkg_rms=bkg_rms
--- a/tests/module/test_ngmix_weight_validation.py
+++ b/tests/module/test_ngmix_weight_validation.py
@@ -41,6 +41,7 @@ import numpy.testing as npt
 import pytest
 
 from shapepipe.modules.ngmix_package.ngmix import (
+    PSF_NOISE,
     Postage_stamp,
     do_ngmix_metacal,
     get_prior,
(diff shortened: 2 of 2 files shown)


cailmdaley:
This is like 5 lines of code ignoring comments and tests, so I'm going to merge without external review.

PR #797, opened 2026-07-09
Coverage nexp masks

### LINKED ISSUE #700: [NEW FEATURE] Create coverage masks for UNIONS
A coverage mask is a required product to go along the UNIONS shape catalogue. It encodes the coverage by exposures, in healsparse format. It can be queried with ra, dec positions to return the number of exposures at that sky position.

Scripts and jupyer notebooks exist, courtesy Mike Hudson. They need to be included as library classes in this PR. Executable python scripts needs to be created on installation from pyproject.toml.

### LINKED ISSUE #836: Build ShapePipe per-CCD failure mask
Build a per-CCD map of where shape measurement failed (PSF-model failure, etc.): ShapePipe records attempted/passed CCDs at run end, then a healsparse mask is built from CCD WCS corners as a pipeline step. Mike has code for this. Possibly (partially) closed by #797 (coverage nexp masks). Open question: confirm whether the observed half-CCD-sized failure chunks are a genuine sub-CCD failure mode or an artifact of how the catalogue was inspected.

Context: [Paris meeting notes](https://unions.skysurvey.cc/en/HostedScienceProjects/2026_Tomographic_Weak_Lensing_Analysis/meetings/20260709_inperson_paris)

martinkilbinger:
## Summary

Implements creation of coverage healsparse masks encoding the exposures on which PSF information is avialable.

The current code, implemented in `utilities` does the following steps:
1. Read CCD list and download corresponding FITS exposure images headers (`header_downloader.py`)
2. Extract CCD corner coordinates using WCS from header (`field_corners_extractor.py`)
3. Create healsparse mask (`coverage_map_builder.py`)

This can be integrated into a workflow. We also need to implement obtaining the list of valid CCDs with PSF model.

## Reviewer Checklist

> Reviewers should tick the following boxes before approving and merging the PR.

- [ ] The PR targets the `develop` branch
- [ ] The PR is assigned to the developer
- [ ] The PR has appropriate labels
- [ ] The PR is included in appropriate projects and/or milestones
- [ ] The PR includes a clear description of the proposed changes
- [ ] If the PR addresses an open issue the description includes "closes #<ISSUE NUMBER>"
- [ ] The code and documentation style match the current standards
- [ ] Documentation has been added/updated consistently with the code
- [ ] All CI tests are passing
- [ ] API docs have been built and checked at least once (if relevant)
- [ ] All changed files have been checked and comments provided to the developer
- [ ] All of the reviewer's comments have been satisfactorily addressed by the developer

Closes #836.


pyproject.toml
```diff
@@ -75,14 +81,20 @@ test = [
...
 summary_run         = "shapepipe.summary_run:main"
 canfar_submit_job   = "shapepipe.canfar_run:run_job"
 canfar_monitor      = "shapepipe.canfar_run:run_log"
 canfar_monitor_log  = "shapepipe.canfar_run:run_monitor_log"
+get_ccd_with_psfs   = "shapepipe.get_ccds_run:run_ccd_psf_handler"
+download_headers    = "shapepipe.coverage_run:run_download_headers"
+extract_field_corners = "shapepipe.coverage_run:run_extract_corners"
+build_coverage_map  = "shapepipe.coverage_run:run_build_coverage"
+plot_coverage_map   = "shapepipe.coverage_run:run_plot_coverage"
+coverage_pipeline   = "shapepipe.coverage_run:run_pipeline"
```
cailmdaley: **Blocker.** `coverage_pipeline` points at `coverage_run:run_pipeline`, which no longer exists (removed in an earlier commit). The installed console script raises `AttributeError` on invocation — drop this line, or restore an orchestrator function.

src/shapepipe/utilities/ccd_psf_handler.py
```diff
@@ -0,0 +1,378 @@
...
+        # Get missing CCDs
+        print("=== get missing CCDs ===")
+        exp_shdu_missing_all = self.get_exp_shdu_missing(patches)
+
+        # Get all exposures used in tiles
+        print("=== get exposures ===")
+        exp_all = self.get_exp(patches)
+
+        # Turn exposures into exposure-single-HDU names (CCDs)
+        exp_shdu_all = summary.get_all_shdus(exp_all, n_CCD)
```
cailmdaley: The missing-CCD set is computed at line 226 but never subtracted here, so this returns every CCD as valid — unfiltered. (The v1.3 path does filter on non-nan FWHM, so the two versions disagree.) Likely part of the valid-CCD-with-PSF work still to come; flagging since the current path silently returns unfiltered. Fix is `exp_shdu_all -= exp_shdu_missing_all` before returning.

src/shapepipe/utilities/coverage_map_builder.py
```diff
@@ -0,0 +1,327 @@
...
+        m = hsp.HealSparseMap.make_empty(nside_coverage, nside, np.uint16)
+
+        # Add polygons for each exposure
+        if verbose:
+            print("Adding polygons to map")
+
+        for i, r in enumerate(rows):
+            ra = [r["tlra"], r["trra"], r["brra"], r["blra"]]
+            dec = [r["tldec"], r["trdec"], r["brdec"], r["bldec"]]
+            poly = hsp.Polygon(ra=ra, dec=dec, value=1)
```
cailmdaley: No RA-wrap or pole guard. Corners arrive straight from astropy as degrees in [0, 360), so an exposure straddling RA=0 yields non-monotonic corners (e.g. `[359.8, 0.2, ...]`) and healsparse fills the ~360° complement, corrupting `nexp` across that RA stripe; near Dec ≈ +90 (the driver runs `dechi=90`) the four-corner ordering also degenerates. Worth unwrapping RA to a common branch per polygon and guarding the pole.

src/shapepipe/utilities/coverage_map_builder.py
```diff
@@ -0,0 +1,327 @@
...
+Author: Mike Hudson, Martin Kilbinger <martin.kilbinger@cea.fr>
+
+"""
+
+import sys
+from os.path import exists
+
+import numpy as np
+import healsparse as hsp
+import hpgeom as hpg
```
cailmdaley, line 19: `hpgeom` is imported directly but isn't in `dependencies` — it only resolves today because `healsparse` pulls it in transitively. Worth declaring explicitly so it doesn't break the day healsparse drops it.

src/shapepipe/utilities/header_downloader.py
```diff
@@ -0,0 +1,281 @@
...
+                    # Create symbolic link
+                    os.symlink(link_source, dest)
+                    return True
+                except Exception as e:
+                    print(f"Could not create symlink from {link_source}: {e}")
+                    # Fall through to download if symlink fails
+
+        # Download from VOSpace
+        try:
+            client.copy(source, dest, head=True)
```
cailmdaley: `client.copy` writes straight to `dest`, so an interrupted transfer leaves a partial `.txt` that the `exists(dest)` check at line 194 treats as complete on the next run — a resumed job then builds on a truncated header. Download to a temp path and `os.rename` on success (atomic).

scripts/sh/build_and_plot_coverage_maps.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
cailmdaley, line 1: No `set -euo pipefail` — a failed build/plot for one version doesn't stop the loop, and the final success message prints unconditionally.

docs/source/pipeline_canfar.md
```diff
@@ -514,11 +515,18 @@ extract_field_corners -i headers_v1.6 -v
 
 Then, build the healsparse coverage mask file as
 ```bash
-build_coverage_map
+build_coverage_map -i exp_ra_dec_v1.6 -o coverage_v1.6.x.hsp -c 128 -n 131072 -v
```
cailmdaley: `-i exp_ra_dec_v1.6`, but the `extract_field_corners` step above is run without `-o`, so it writes the default `exp_ra_dec.txt` — this filename is never produced, and the example throws `FileNotFoundError` if followed verbatim. Align the two.

cailmdaley (commented):
Thanks Martin. The merge conflict's resolved — I merged `develop` in (the branch had fallen well behind), reconciled `pyproject.toml`, and regenerated `uv.lock`, so it's `MERGEABLE` again. Structure looks clean; the main thing is a scope question, then a few specific items inline.

**Per-exposure vs per-CCD coverage.** The mask is currently built per *exposure*: `get_exposures()` collapses the per-CCD valid-PSF IDs down to unique exposure numbers, `_megacam_field_corners()` draws a single polygon over the whole ~1° MegaCam mosaic, and each exposure adds a flat +1 across it. So an exposure with even one valid-PSF CCD stamps +1 over its entire field — including the sky where its *other* CCDs had no valid PSF. `nexp` then over-reports PSF depth exactly where CCD-level failures cluster (chip gaps, flagged/edge CCDs). The per-CCD validity is already computed upstream, so the fix is to carry the `exp-shdu` identity through and draw one polygon per surviving CCD instead of collapsing to exposures.

Two questions so we're reviewing against the right target: (1) do you intend to build the mask at CCD level? (2) does the valid-CCD-with-PSF selection land in this PR, or a follow-up? The description flags it as still-to-do, so I want to make sure we're not measuring against the wrong scope.

I think this wants test coverage before it merges — there's none yet, and the pure parsing functions (exposure-number regex, header→WCS split, the corner convention) are exactly where silent failures would hide, and are cheap to cover. Happy to write those with you, and to help wire the per-CCD path if that's the direction.

A few specific items inline.

— Fable, on behalf of Cail

cailmdaley:
Update — dug into the failing checks and took care of them so this is green for you.

Merging `develop` in ran its full suite against the branch, surfacing that all six coverage console scripts failed the entry-point smoke test (`tests/unit/test_entrypoints.py::test_console_entrypoint_help_runs`). Two causes, both fixed:

- The five runners parsed the full `sys.argv` (including `argv[0]`) rather than `sys.argv[1:]`, so `-h` collided with the integer `-c` option (`error: option -c: invalid integer value: '-h'`). Fixed to slice `argv[1:]`, matching `cs_util.args.parse_options`' own convention.
- `coverage_pipeline` pointed at `run_pipeline`, which had been removed — dropped the dangling entry point (easy to re-add as a real orchestrator later if you'd like one).

Also declared `hpgeom` explicitly (imported directly by `coverage_map_builder`, previously only resolving transitively via `healsparse`).

And I added `tests/module/test_coverage.py` — unit + property tests for the pure logic: exposure-number parsing, the CCD-list → unique-exposure reduction, the MegaCam field-corner convention, the multi-HDU header → WCS split, power-of-two `nside` validation, and a regression guard for the `-h` handling. Happy to extend these as the feature grows.

CI is green now (357 passed). With that, the one substantive thing still open is the design question at the top: whether the mask should be built per-CCD — so PSF validity is honoured per-CCD footprint — rather than per-exposure. That one's yours; glad to help wire it once you've settled the intent (and whether the valid-CCD-with-PSF selection lands here or in a follow-up).

— Fable, on behalf of Cail

pyproject.toml
cailmdaley: Done — dead `coverage_pipeline` entry point dropped in 8b9387b3.

src/shapepipe/utilities/coverage_map_builder.py
cailmdaley: Done — `hpgeom` declared in `dependencies` (8b9387b3).

martinkilbinger:
Update on workflow, the 0th step is:

0. Get CCDs with valid PSF: `get_ccds_with_psf` using `CcdPsfHandler` (needs updating for `v2.0`)

And `field_corner_extractor.py` is indeed not doing the CCD-wise masking, but only looking at the entire MegaCam field, which is wrong. This needs fixing.

cailmdaley:
want me to take care of these issues @martinkilbinger ?

martinkilbinger:
> want me to take care of these issues @martinkilbinger ?

yes please :-)

There was an earlier question, not sure whether this has been answered, but I confirm that here we want single-CCD-based masks, to account for available/missing PSF. Exposure-count-based masks, or coverage masks, are done elsewhere.

cailmdaley:
Done — the branch now builds single-CCD-based masks. The chain:

1. **`get_ccds_with_psf`** → valid-PSF CCD IDs. The v ≥ 1.4 path now actually subtracts the missing-CCD set (it previously returned every CCD as valid), and `-V v2.0` is accepted (see below).

   On "needs updating for v2.0": following your [#812 note that the patch concept is removed in v2.0](https://github.com/CosmoStat/shapepipe/pull/812#discussion_r3577047406), `-V v2.0` is now patch-less — the handler reads `summary/missing_job_32_all.txt` and `exp_numbers.txt` relative to the run root instead of iterating `P1..P9` (a new `-p` option restores an explicit `P1..P<n>` layout if ever needed). The job-32 numbering and summary filenames are unchanged on develop, so those are kept as-is — flag if the v2 runs will change the summary conventions themselves.
2. **`extract_field_corners --ccd_list <step-1 output>`** → one row of corners per valid CCD, from each HDU's own WCS (image dims read from `ZNAXIS` — the `.fz` headers' `NAXIS` are the compressed-table dims). This is where the PSF-validity filter lives; without `-l` all CCDs are extracted.
3. **`build_coverage_maps`** → one polygon per CCD, with RA-wrap and pole guards. CCDs of one exposure don't overlap, so pixel values still count exposures with valid PSF.

Robustness along the way: extractor resume is keyed on individual CCD IDs (interrupted runs and expanded CCD lists complete cleanly), and header downloads are atomic, so partial files can't poison a resumed run.

57 tests cover the chain — including the nexp value contract (overlapping CCDs from two exposures → 2 in the overlap, 1 outside) — and `pipeline_canfar.md` documents the full flow with aligned filenames. CI green. Over to you for review.

**Update (2026-07-20):** `get_ccds_with_psf -V` now defaults to the patch-less `v2.0` (`1ffcd4b1`), matching the convention adopted in #812's `collate_star_cat.py`; `v1.x` layouts are selected explicitly with `-V v1.6`. Tests still green (45 passed, in-container).

cailmdaley:
@martinkilbinger please take a look, should be good now!

cailmdaley:
i think this closes #836?

cailmdaley:
@martinkilbinger this is ready for review/sign off!

docs/source/pipeline_canfar.md
```diff
@@ -29,7 +29,15 @@ canfar auth switch default
...
-Set the current patch in the shell as
+> **Patch convention by catalogue version.** Runs up to `v1.6` are organised
+> in sky patches `P1`–`P9` (7 patches for `v1.3`/`v1.4`, 8 for `v1.5`), each
+> patch a run directory of its own with its `summary/` subdirectory. **The
+> patch concept is removed in `v2.0`**: there is a single run root, and
+> summary files (`summary/missing_job_32_all.txt`, `exp_numbers.txt`) live
+> directly under it. Tools that take a catalogue version (e.g.
+> `get_ccds_with_psf -V`) follow this convention automatically.
+
+For a `v1.x` run, set the current patch in the shell as
```
martinkilbinger: I don't think v2.0 will still use the previous summary class and outputs. This will be replaced by more robust and efficient bookkeeping of missing CCDs.

src/shapepipe/utilities/ccd_psf_handler.py
```diff
@@ -93,7 +93,8 @@ def update_params(self):
             n_patch = 7
         elif version == "v1.5":
             n_patch = 8
-        elif version == "v1.6":
+        elif version in ("v1.6", "v2.0"):
+            # v2.0 (ngmix) covers the same 9-patch footprint as v1.6
             n_patch = 9
```
martinkilbinger, line 97: Since the concept of patches is made obsolete in v2.0, this should be replaced by updated handling.

martinkilbinger:
I added comments about patches in v2.0: The concept of patches should be removed for v2.0. The script `get_ccds_with_psf` should be updated or replaced with something to better reflect the new v2.0 processing structure.

cailmdaley:
Thanks Martin, you're right about v2.0: those summary files won't exist. I removed the v2.0 references so this lands as the v1.x tool. For v2.0 we'll open a separate PR on the Snakemake stack once this is in develop: valid-PSF CCDs from the persisted PSF products (#879), corners from the split-exposure headers, reusing the mask builder from here. Does that order work for you?

(Also re-locked `cs_util` to pick up CosmoStat/cs_util#82 — `FootprintPlotter.plot_area` had gone missing in April, so `plot_coverage_map` was raising `AttributeError` on develop.)

— Claude on behalf of Cail

cailmdaley:
Given that Martin's feedback only touched on the v2 question, which deserves its own PR, I'm merging this so it's not blocking downstream work.

PR #802, opened 2026-07-10
Fix nfev_fit int64 dtype lock crashing batch-saved ngmix output (#795)

### LINKED ISSUE #795: [BUG] ngmix crash
@fabianhervaspeters has reported seeing crashs of ngmix after a batch of 1000 object is written to file.

We should check that batch writing and reading is well implemented.

cailmdaley:
## Summary
- `compile_results` filled the `nfev_fit` column with `np.nan` on failed fits — the only int-like output column with a float fallback.
- With `SAVE_BATCH` enabled, `save_results` locks column dtypes from batch 1 (fresh `save_as_fits` write) and appends every later batch into a structured array typed to those locked dtypes. If batch 1 happens to have zero failed fits, `nfev_fit` locks to int64 (FITS `'K'`); the first later batch containing a failed fit then feeds a NaN into that int64 column, raising `ValueError: cannot convert float NaN to integer` at append time — crashing tiles mid-run, exactly at a batch boundary as reported.
- Fix: use ngmix's own `-1` failed/absent-`nfev` sentinel (already documented in the adjacent comment) instead of `np.nan`. `nfev_fit` is now int64 with `-1` meaning failed/absent, dtype-stable across every batch.

## Test plan
- [x] New regression test `test_save_results_batches_survive_a_later_failed_fit` drives the real `save_results` over a two-batch sequence (batch 1 all-success, batch 2 with a failed fit type) — fails on develop with the exact reported crash at `ngmix.py:612`, passes with the fix.
- [x] `tests/module/test_ngmix.py` — 18 passed (in-container).
- [x] `pytest -m "not slow" tests/science` — 11 passed, no regressions.
- [x] Checked every `nfev` reference in the codebase — no consumer reads or NaN-masks the output column.

Fixes #795

### DIFF
--- a/src/shapepipe/modules/ngmix_package/ngmix.py
+++ b/src/shapepipe/modules/ngmix_package/ngmix.py
@@ -477,9 +477,13 @@ class Ngmix(object):
                 )
                 # ngmix 2.x reports the solver's function-evaluation count
                 # (nfev, ~tens-hundreds; -1 on some failures), not the v1
-                # 1-5 retry count, so the column is named accordingly.
+                # 1-5 retry count, so the column is named accordingly. Fits
+                # that fail before nfev is ever set (e.g. the NaN-filled
+                # branch above) fall back to the same -1 sentinel, keeping
+                # the column int64 (FITS 'K') across every batch: nfev_fit
+                # is int64 with -1 meaning failed/absent.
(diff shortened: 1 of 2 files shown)


cailmdaley:
LGTM

Issue #767, opened 2026-06-29
ngmix centroid: switched to WCS in #741 — confirm correct, and any reason to keep HSM?


cailmdaley:
Following Fabian's recommendation, #741 switches the ngmix metacal centroid from HSM adaptive-moment re-centering to the WCS (the catalog sky position projected through the WCS). Opening this to confirm with the group that the WCS centroid is the right choice, and to ask whether there's any reason to keep the HSM approach around before we remove that code path entirely (a separate, non-blocking PR).

For context — reassuringly, this is not a v2.0 regression: HSM moment-centroiding is long-standing inherited behavior, first introduced in 2020 (`bde89f5f`) and carried through the 2023 module overhaul (`441797e2`) into both `develop` and v2.0; the recent v2.0 work only made the centroid source *selectable* (`0ee0e944`). `develop` currently uses HSM unconditionally — the WCS option arrives with #741.

cc @aguinot

— Claude on behalf of Cail

aguinot:
Hey,
Let's make a little clarification, what is called WCS centroid is actually the coadd centroid. I think it would be better to call it this to make more explicit where it comes from.
So, this centroid is actually the one used to get the stamps in the first place and I think it would be good to see how this is done:
- Coadd x,y locations are transformed to ra, dec
- ra, dec are transformed to single exposures x, y
- Those x, y are rounded in `sf_tools`
- A stamp is extracted around the rounded locations with a fixed size (meaning that there is padding if the object is close to an edge)

Using the coadd centroids is fine as long as the locations are computed in the exact same way as in `sf_tools`. To do that, there are several options, some cleaner than others:
- Looking into `sf_tools` and copy/paste how the rounding is done and be careful to handle the padding correctly
- When extracting the stamp with `sf_tools`, also extract the x, y location and more importantly the rounded part
- Stop using `sf_tools` (I am not sure if Sam still support the library so it might be good to switch), re-implement the stamp extraction and implement the propagation of the locations at the same time

Using the HSM centroid by-pass all this. It gives a pretty good prior and unlikely to fail (but can still fail in some cases, might be better to do it within the mcal loop). For very faint objects it might be tricky to get a prior. Using weighted moments would be significantly better than using adaptive moments because they cannot fail. The HSM centroid is not the only thing that get used though, it also gives a prior on the size and fluxe which can be very useful to make the fit faster and more stable. That being said, they are probably overwritten by the ngmix prior..

fabianhervaspeters:
[...]

PR #804, opened 2026-07-10
Breakdown grid v2: paired six-arm estimator, honest bootstrap errors, degeneracy guards

cailmdaley:
## Summary

Replaces the v1 breakdown-grid harness (16 unpaired seeds/cell, no uncertainties) with a statistically sound v2:

- **Paired ±γ arms on identical noise realizations** — six arms per seed (g1±, g2±, psf±) sharing the RNG stream, so measurement noise cancels in the branch difference (same statistical move as #796 at survey scale).
- **Joint seed-resample bootstrap** on every reported quantity — one index array per replicate feeds numerators, R denominators, and the legacy estimator, so the pairing survives the resampling.
- **Degeneracy guard**: cells where |R̄| < 5·se(R̄) report null + `degenerate: true` instead of dividing (the v1 grid printed m2 = −58.6 at the point-source/low-S/N corner from exactly this division).
- Independent seeds per cell (v1 recycled one seed set across all 24 cells, correlating the apparent "trends"), per-band seed counts {high 100, moderate 400, low 1600}, per-cell npz of raw per-(seed,arm) records so reanalysis never refits, `--true-noise` ablation knob (routes `prepare_ngmix_weights` onto the per-pixel true-inverse-variance path via `stamp.bkg_rms`).
- Plotter: error bars on every point, degenerate cells as open markers, ±5e-3 tolerance band.

## Why it matters

The v1 panel showed alarming S/N-dependent m (up to +0.31 at S/N≈50) and c trends. The v2 grid (24 cells, ~100k metacal fits, SLURM) shows they were statistical artifacts of the v1 harness: **every resolved cell in every band is consistent with m = 0** (precision ~1e-3 high band, ~1e-2 moderate, ~1.5e-2 low), **|c1| < 1e-3 and leakage α ≤ 0.2% everywhere**, and R11 ≈ R22 on resolved rungs. The live `do_ngmix_metacal` path is healthy at survey-relevant S/N; the earlier panel was measuring its own noise floor.

## Test plan
- [x] Pairing identity: noise draw + sub-pixel shift bit-identical across all six arms of a seed
- [x] Bootstrap validated against direct per-seed error decomposition (numerator 0.0087 ⊕ R̄-noise 0.0264, corr +0.42 → observed 0.0243 at the 80-seed pilot cell)
- [x] Degeneracy guard fires at (r=0.15, low) instead of printing an exploded number
- [x] Full 24-cell production run completed (job 808255); results in the twin results tree
- [x] No `src/` changes — harness-side only (`scripts/python/`)

### DIFF
--- a/scripts/python/plot_breakdown_grid.py
+++ b/scripts/python/plot_breakdown_grid.py
@@ -2,17 +2,20 @@
 # requires-python = ">=3.10"
 # dependencies = ["numpy", "matplotlib", "seaborn"]
 # ///
-"""Column figure for the resolution x noise breakdown grid.
+"""Column figure for the resolution x noise breakdown grid (v2).
 
-Reads the JSON emitted by ``run_breakdown_grid.py`` and draws three vertically
-stacked panels sharing the x-axis -- multiplicative bias ``m``, metacal response
-``R11``, and additive bias ``|c1|`` vs the resolution ratio ``r`` -- with the
-three S/N bands as hue. The ``m`` panel is linear with a widened y-range so the
-moderate/low-S/N climb stays on-plot while the +/-5e-3 tolerance band is still
-drawn. Seaborn is not in the ShapePipe sif, so this runs on the host:
+Reads the JSON emitted by ``run_breakdown_grid.py`` (v2 schema) and draws three
+vertically stacked panels sharing the x-axis -- multiplicative bias ``m1`` and
+``m2``, and additive bias ``|c1|`` vs the resolution ratio ``r`` -- with the
+three S/N bands as hue. **Every point carries its bootstrap error bar**; that is
+the whole point of v2. Cells whose response was consistent with zero (the
+degeneracy guard fired, ``m`` null) are drawn as open markers at 0 with a hatch
+so a reader never mistakes "we refused to divide" for "m = 0". The ``m`` panels
+keep the ``+/-5e-3`` tolerance band drawn as a sliver at 0. Seaborn is not in the
+ShapePipe sif, so this runs on the host:
 
     uv run scripts/python/plot_breakdown_grid.py \\
-        --input <run>/results/baseline/breakdown_grid/breakdown_grid.json \\
+        --input <run>/results/baseline/breakdown_grid/breakdown_grid_v2.json \\
         --output <figures>/breakdown_grid.png
 """
 import argparse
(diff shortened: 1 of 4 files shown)


cailmdaley:
**Update: S4 production-parity ablations — the v2 PASS survives production settings.**

Three new parity knobs on the harness (`--centroid-source hsm|wcs`, `--wcs-g1/--wcs-g2/--wcs-theta-deg`, existing `--n-epochs`), five full-grid ablation arms with the same seeds as the baseline (jobs 808257–808262):

| arm | result |
|---|---|
| `wcs` centroid (production default since `31ae736c`) | paired per-seed Δm₁ vs `hsm`: **few ×1e-6** at S/N≈1000; \|Δm₁\| < ~1e-3 at every resolved rung down to S/N≈15 — the centroid source is not an m-lever (relevant to #767) |
| WCS jacobian shear g1=0.1 (the esheldon/ngmix#72 config that produced m≈−0.2) | no effect: \|Δm\| < 1.5e-3 (1σ, high band) — **the ngmix#72 pathology is absent from our stack**, at ~5× any realistic MegaCam distortion |
| realistic jacobian (g1=0.02, g2=0.01, θ=10°) | no effect (max \|Δ\|/σ = 0.57) |
| n_epochs=2 | the old single-seed CI ladder's −2e-3 does **not** reappear: every resolved high-band cell within ~1e-3 of zero |
| production combo (wcs centroid + realistic jacobian + 2 epochs + true-noise weights) | all cells within tolerance; max resolved high-band \|m₁\| = 1.0e-3 ± 0.5e-3 |

No \|Δ\|/σ ≥ 2 anywhere across ~300 arm×band×cell comparisons (the 1.7–1.9σ cluster in the two 2-epoch arms traces to a shared baseline fluctuation — both arms sit *closer* to zero than the baseline at that cell). Comparison figure + per-arm significance tables: `plot_s4_ablations.py`, with a `--paired` mode that cancels shape noise for bit-identical-image arms (that's what resolves the centroid effect at 1e-6).

Two notes on the diff since the original description: the test-plan line "no `src/` changes" predates these commits — `simulate.py` gained two backward-compatible kwargs (all 14 `make_data` call sites verified unaffected; legacy path smoke-verified bit-identical against the baseline npz). And a robustness fix: a single `BootPSFFailure` in ~1e5 fits used to kill the whole grid; it's now recorded as a flagged object (production behaviour), with the rate surfaced in `fail_frac`.

PR #811, opened 2026-07-12
Cut postage stamps in-house and reuse their exact centroid in ngmix (drop sf_tools)

### LINKED ISSUE #767: ngmix centroid: switched to WCS in #741 — confirm correct, and any reason to keep HSM?

cailmdaley:
Closes #767.

ShapePipe cuts a small "postage stamp" image around every object before measuring its shape with ngmix. Until now the cutting was done by `sf_tools` (unmaintained for years), and ngmix independently re-computed the object's position — through a *different* WCS library — to place its centroid prior. Two computations of the same number can disagree: near a rounding tie, a milli-pixel WCS difference rounds to a different pixel and puts the prior a full pixel off the stamp it was cut from (#767).

## What changes

- **In-house cutter.** `vignetmaker.get_stamps()` replaces `sf_tools.FetchStamps`: zero-pad by the stamp radius, slice around `round(pos)`. Bit-identical to sf_tools for in-bounds objects (tested directly against it, plus a hand-computed golden test that keeps the behavior pinned once sf_tools is uninstalled). An out-of-bounds centre now raises instead of being silently wrapped modulo the image shape.
- **One centroid, shared.** The cutter records the rounded pixel (`INT_POS`) and sub-pixel offset (`OFFSET`) alongside each vignette in the same sqlite entry — no sidecar file. ngmix's `wcs` centroid consumes the stored `OFFSET` instead of re-projecting and re-rounding: one projection, one rounding, so the stamp and the centroid prior cannot disagree. This is Axel's condition on #767, met by construction.
- **`CENTROID_SOURCE hsm|wcs` stays selectable**, so the eventual real-data hsm-vs-wcs comparison runs against a single pipeline.
- **`sf_tools` dropped** from `pyproject.toml`, `uv.lock`, and the module `depends`.

Note: vignette files produced before this change carry no `OFFSET`; running new ngmix on them with `CENTROID_SOURCE wcs` fails fast with a message saying to regenerate the vignettes. Intended — a propagated offset genuinely doesn't exist for old files.

**Out of scope:** the real-data hsm-vs-wcs A/B (needs a real-data run), and removing the HSM path (the open question on #767).

— Claude on behalf of Cail


cailmdaley:
@fabianhervaspeters please take a look at this!

PR #698, opened 2025-05-12
fourth-order moments

martinkilbinger:
## Summary

Implementing fourth-order moments.

Solves https://github.com/CosmoStat/shapepipe/issues/697

## Reviewer Checklist

> Reviewers should tick the following boxes before approving and merging the PR.

- [ ] The PR targets the `develop` branch
- [ ] The PR is assigned to the developer
- [ ] The PR has appropriate labels
- [ ] The PR is included in appropriate projects and/or milestones
- [ ] The PR includes a clear description of the proposed changes
- [ ] If the PR addresses an open issue the description includes "closes #<ISSUE NUMBER>"
- [ ] The code and documentation style match the current standards
- [ ] Documentation has been added/updated consistently with the code
- [ ] All CI tests are passing
- [ ] API docs have been built and checked at least once (if relevant)
- [ ] All changed files have been checked and comments provided to the developer
- [ ] All of the reviewer's comments have been satisfactorily addressed by the developer

Issue #697, opened 2025-05-12
[NEW FEATURE] Fourth-order moments


martinkilbinger:
Implement fourth-order moments for PSF diagnostics.

@sachaguer @martinkilbinger 

Resolved by https://github.com/CosmoStat/shapepipe/pull/698

PR #812, opened 2026-07-12
HSM: measure PSF/star shapes directly in sky coords (use_sky_coords)

### LINKED ISSUE #775: HSM: measure PSF/star shapes directly in sky coords (use_sky_coords), drop the WCS re-transform
galsim's HSM `FindAdaptiveMom` now accepts `use_sky_coords=True` to return shapes in sky coordinates at the measurement site, making the separate WCS-rotation step (`scripts/python/convert_psf_pix2world.py`) redundant. Axel endorsed this on #741 (2026-06-22). Verified `use_sky_coords` is **not** in #761/#741/develop, so it's genuinely still to do. Small self-contained PR: covers PSFEx + MCCD; the WCS knowledge moves *into* the HSM call (the rotation script is a live CANFAR step, not a pure deletion); validate world-frame `*_PSF_HSM`/`*_STAR_HSM` columns identical to today. Rebases onto the merged #761 grammar (same `*_PSF_HSM` columns) — do **after the #741 merge**. Longer-term (not blocking the v2.0 merge). Implementation spec: the held shuttle constitution; PR attaches here via Development when ready. — Claude on behalf of Cail

cailmdaley:
Closes #775.

ShapePipe measured PSF/star HSM adaptive moments in the pixel frame, then rotated the ellipticities into sky coordinates in a separate post-processing step (`scripts/python/convert_psf_pix2world.py`, a WCS-Jacobian decomposition). galsim's `FindAdaptiveMom` now does this at the point of measurement via `use_sky_coords=True` (available since galsim 2.5; we pin ≥2.8) — per @aguinot's suggestion on #741. This PR moves the WCS knowledge into the HSM call and drops the redundant rotation.

## What changed

- **Direct sky-frame measurement.** Every PSF/star `FindAdaptiveMom` call site — PSFEx single-exposure and multi-epoch (`psfex_interp.py`), MCCD interpolation (`mccd_interpolation_script.py`, `shapepipe_auxiliary_mccd.py`) — now attaches the local WCS (linearised at the object's image position) to the stamp and passes `use_sky_coords=True`. Multi-epoch paths use the per-CCD WCS log already in scope; single-exposure paths read the WCS from the galcat's `LDAC_IMHEAD` header.
- **Rotation removed where redundant; script renamed.** `convert_psf_pix2world.py` no longer rotates the in-repo `E1/E2/SIGMA_*_HSM` shapes. Per review, it is renamed to `collate_star_cat.py` to reflect its remaining purpose: collate the per-exposure PSF validation catalogues into star catalogues (positions, MCCD CCD id, merge).
- **Patch-less `v2.0` support.** The script takes `-V` (default `v2.0`), following the `get_ccds_with_psf` convention from #797. the patch-less default reads from the single run root and writes `validation_psf_conv-<idx>.fits` (no patch token; downstream `combine_runs.bash` / `config_Ms_psfex_conv.ini` globs match both schemes). `-V v1.6 -P …` selects the unchanged v1.x behaviour. Invalid versions raise. The collate section of `pipeline_canfar.md` shows both invocations.
- **Exception kept deliberately:** the MCCD *fit-validation* `PSF_MOM_LIST`/`STAR_MOM_LIST` columns come from the external `mccd` library, which still measures in the pixel frame — that one branch keeps its rotation (with a docstring caveat) until `mccd` itself adopts `use_sky_coords`.
- **Equivalence proven in-tree.** `tests/module/test_hsm_sky_coords.py`: a synthetic Gaussian stamp under a nontrivial local WCS (0.187″/pix, 28° rotation, shear), measured pixel-frame + old `transform_shape` math vs `use_sky_coords=True` — g1/g2/sigma agree to rtol 1e-6. Plus `test_collate_star_cat.py` pinning the v1.x/v2.0 path and filename logic. All 14 tests pass in the develop container.

## Notes for review

- **Sigma unit change (intended):** `moments_sigma` now comes back in world units (arcsec) rather than pixels, so `HSM_T_* = 2σ²` is arcsec² — which is what the old pipeline produced *after* the rotation step, so end-of-pipeline catalogues are unchanged. Downstream consumers (`merge_starcat`, ngmix T-guess) checked: no pixel-scale assumption.
- **No double rotation:** grep-verified nothing downstream applies a second Jacobian/`getDecomposition` to the HSM ellipticities; `merge_starcat` passes them through.
- A subtle off-by-one (galsim's FITS 1-indexing vs astropy origin-0 positions) in the local-WCS evaluation point was caught in internal review and fixed (`f153ea28`).

An independent adversarial review of the full diff ran before opening this PR; both findings it raised (the external-mccd branch above, and the indexing fix) are addressed.

— Claude on behalf of Cail

https://claude.ai/code/session_01Gao7Pi5QDBgnTuagV7HT2w

### DIFF
--- a/docs/source/pipeline_canfar.md
+++ b/docs/source/pipeline_canfar.md
@@ -333,27 +333,43 @@ shapepipe_run -c $SP_CONFIG/config_Ms_$psf.ini
 shapepipe_run -c $SP_CONFIG/config_Pl_$psf.ini
 ```
 
-#### Collate star catalogues per patch
+#### Collate star catalogues
 
-Collate all input validation PSF files and create directories per patch `P?`,
-producing files `validation_psf_conv-<patchnum>-<idx>.fits` (for the v1.4 setup
-only one file). This gathers positions (X/Y/RA/DEC) and the MCCD CCD id.
+Collate all input validation PSF files into star catalogues, gathering positions
+(X/Y/RA/DEC) and the MCCD CCD id.
 
 Note: HSM shapes are no longer rotated into world coordinates at this step. The
 PSF/star ellipticities and sizes are now measured directly in sky coordinates
 during PSF interpolation (galsim `FindAdaptiveMom(use_sky_coords=True)`), so
-`convert_psf_pix2world.py` only collates and passes the shapes through.
+`collate_star_cat.py` only collates and passes the shapes through.
+
+> **v2.0 is patch-less.** Runs up to `v1.6` are organised in sky patches
+> `P1`..`P<n>`, each patch a run directory. `v2.0` removes the patch concept: a
+> single run root, outputs directly under it. `v2.0` is the default; select an
+> older layout with `-V` (e.g. `-V v1.6`). For `v2.0` the patch loop and the
+> `-P` option no longer apply, and the patch token drops from the output
+> filename (`validation_psf_conv-<idx>.fits` instead of
+> `validation_psf_conv-<patchnum>-<idx>.fits`).
 
 ```bash
 cd /path/to/version
-mkdir stat_car
+mkdir star_cat
 cd star_cat
 ```
 
-For each patch run
+For `v2.0` (the default), run once against the patch-less run root, producing
+files `validation_psf_conv-<idx>.fits`:
 
 ```bash
-convert_psf_pix2world.py -i .. -P $patchnum -v
+collate_star_cat.py -i .. -v
+```
+
+For `v1.x`, pass the version explicitly and run once per patch, creating a
+directory per patch `P?` and producing files
+`validation_psf_conv-<patchnum>-<idx>.fits` (for the v1.4 setup only one file):
+
+```bash
+collate_star_cat.py -i .. -V v1.6 -P $patchnum -v
 ```
 
 Combine previously created files as links within one ShapePipe run directory (for the v1.4 setup only one link).
--- a/docs/source/pipeline_canfar.md
+++ b/docs/source/pipeline_canfar.md
@@ -333,10 +333,16 @@ shapepipe_run -c $SP_CONFIG/config_Ms_$psf.ini
 shapepipe_run -c $SP_CONFIG/config_Pl_$psf.ini
 ```
 
-#### Convert star catalogue to wCS
+#### Collate star catalogues per patch
 
-Convert all input validation PSF files and create directories per patch `P?`.
-Create files `validation_psf_conv-<patchnum>-<idx>.fits` (for the v1.4 setup only one file):
+Collate all input validation PSF files and create directories per patch `P?`,
+producing files `validation_psf_conv-<patchnum>-<idx>.fits` (for the v1.4 setup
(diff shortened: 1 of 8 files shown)


martinkilbinger:
We should revisit `convert_psf_pix2world.py`. In `v2.0` we will no longer define patches. Renaming or re-scoping this script should reflect its main purpose if it is still needed: to merge the star catalogues. If pixel to WCS is needed for MCCD, this should just be implemented in the MCCD library, we can check in with @tobias-liaudat .

docs/source/pipeline_canfar.md
```diff
@@ -333,10 +333,16 @@ shapepipe_run -c $SP_CONFIG/config_Ms_$psf.ini
...
-Convert all input validation PSF files and create directories per patch `P?`.
-Create files `validation_psf_conv-<patchnum>-<idx>.fits` (for the v1.4 setup only one file):
+Collate all input validation PSF files and create directories per patch `P?`,
+producing files `validation_psf_conv-<patchnum>-<idx>.fits` (for the v1.4 setup
+only one file). This gathers positions (X/Y/RA/DEC) and the MCCD CCD id.
+
+Note: HSM shapes are no longer rotated into world coordinates at this step. The
+PSF/star ellipticities and sizes are now measured directly in sky coordinates
+during PSF interpolation (galsim `FindAdaptiveMom(use_sky_coords=True)`), so
+`convert_psf_pix2world.py` only collates and passes the shapes through.
```
martinkilbinger, line 353: Update to reflect removal of the patch concept in `v2.0`

martinkilbinger (changes requested):

cailmdaley:
Thanks Martin — both points addressed.

- `convert_psf_pix2world.py` is renamed to `collate_star_cat.py`, re-scoped to its remaining purpose: collating the per-exposure PSF validation catalogues into star catalogues. No shape rotation happens there any more, with one exception: the external-mccd `PSF_MOM_LIST`/`STAR_MOM_LIST` columns are still measured in the pixel frame upstream, so that one branch keeps the rotation until `mccd` adopts `use_sky_coords` — agreed that belongs in the MCCD library; we can raise it with Tobias separately.
- The script now takes `-V` following the same convention as `get_ccds_with_psf` in #797, defaulting to the patch-less `v2.0` (single run root); v1.x runs pass `-V v1.6 -P …` and behave as before. The collate section of `pipeline_canfar.md` shows both.

— Fable on behalf of Cail

aguinot:
Hello,
While you're looking into this, it might be good to add higher order moments. It has been very useful in other surveys.

cailmdaley:
Good point — agreed, and it turns out this was attempted before: #698 (fourth-order moments, PSFHOME-style) was closed unmerged and is now stale. It edits exactly the two functions this PR rewrites. We've reopened #697/#698 to track it and will implement as a follow-up on top of this PR: the spin-2 fourth-moment combinations measured in the sky frame (which removes the pixel→sky rotation the old code needed), plus galsim's spin-0 `moments_rho4`, with tests for both.

— Fable on behalf of Cail

martinkilbinger (approved):
LGTM!

PR #818, opened 2026-07-13
ci: adopt ruff, drop pytest-pycodestyle/pytest-pydocstyle

### LINKED ISSUE #791: Adopt ruff linting (would have caught the un-raised-exception pattern)
## Motivation

PR #789 (fixing #785) found the same bug — an exception **constructed but never raised**, silently falling through an error path — at **five independent sites** across the codebase (`psfex_interp_runner.py`, `mccd_interp_runner.py` ×2, `mask_package/mask.py`, `pipeline/file_handler.py`, `pipeline/file_io.py`). Ruff's `PLW0133` (useless-exception-statement) flags exactly this pattern; a linter in CI would have caught all five at introduction time.

## Proposal

- Add `ruff` to the dev tooling with a conservative initial rule set (e.g. `E`, `F`, `PLW0133` and friends) — start narrow, expand as the codebase is cleaned.
- Wire it into CI so new violations block, without requiring a big-bang cleanup of existing code (per-file ignores or `--diff`-scoped enforcement).

Sibling precedent: sp_validation has a project-wide ruff pass planned (CosmoStat/sp_validation#204).

— Claude (Fable) on behalf of Cail

cailmdaley:
Part of #791.

## Why

pytest **9.1** removed the deprecated `path` argument from the `pytest_collect_file` hookspec. The unmaintained **`pytest-pydocstyle`** plugin still declares `path` in its hookimpl, so pytest aborts at plugin *registration* — before collecting a single test:

```
pluggy._manager.PluginValidationError: Plugin 'pydocstyle' for hook 'pytest_collect_file'
Argument(s) {'path'} are declared in the hookimpl but can not be found in the hookspec
```

This turns CI red on **every** dependency PR that bumps pytest to 9.1+ — currently Renovate #816 (all-non-major) and #817 (lock maintenance), both of which fail here and nowhere else (the image builds fine).

Crucially, these plugins were **dead weight**: never enabled via flags, never configured, enforcing nothing. They only broke CI because merely installing them registers the incompatible hook.

## What

- Replace the entire style-tool stack — `black`, `isort`, `pytest-pycodestyle`, `pytest-pydocstyle` — with **ruff**, one maintained tool that subsumes all of them: pycodestyle (`E`/`W`), Pyflakes (`F`), import sorting (`I`, replaces isort), numpydoc docstrings (`D`). `ruff format` replaces `black`.
- **Bump pytest 9.0.3 → 9.1.1 in `uv.lock`**, so this PR's own CI exercises the exact scenario it fixes — **green CI here is the proof** — and completes the bump Renovate was attempting.
- `uv.lock` regenerated (the image installs via `uv sync --frozen`, so the lock must match pyproject): removes black, isort, pycodestyle, pydocstyle, both pytest plugins; adds ruff; bumps pytest.

## Scope note — no CI gate yet

Enforcement is **deliberately not wired into CI in this PR** — parity with today, where no style gate existed. A clean `ruff check` currently reports ~600 findings, including some genuine bugs (undefined names, unused variables, bare `except`) that deserve a dedicated follow-up rather than being buried in a dependency-unblock PR. This PR's job is to unblock pytest 9.1 and land the maintained ruff foundation; enforcement can be turned on incrementally.

Once merged, #816 / #817 rebase clean (their pytest bump is already done here).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,8 +60,7 @@ jupyter = [
     "snakemake>=9.22.0",
 ]
 lint = [
-    "black",
-    "isort"
+    "ruff",
 ]
 release = [
     "build",
@@ -71,8 +70,6 @@ test = [
     "hypothesis>=6.155.2",
     "pytest>=8.3",
     "pytest-cov>=5.0",
-    "pytest-pycodestyle>=2.4",
-    "pytest-pydocstyle>=2.4",
 ]
 fitsio = ["fitsio"]
 dev = ["shapepipe[doc,jupyter,lint,release,test,fitsio]"]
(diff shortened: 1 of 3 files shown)


cailmdaley:
LGTM

Issue #808, opened 2026-07-12
Set up ShapePipe on Nibi and run end-to-end on a first batch of P3 tiles


cailmdaley:
## Goal

Get ShapePipe running end-to-end on Nibi over a **handful of P3 tiles** — the first real run on data. This is deliberately small: the point is to stand up the container, wire the pipeline to our pre-staged data, get a valid shape catalogue out the far end, and **measure how long a tile actually takes and how it parallelizes** — so we can size the full P3 run realistically. Second issue on the **First run on data (Nibi)** milestone, following #801.

## What's already in place (from #801)

P3 data is staged locally in `/project/def-mjhudson/unions-wl/` (see its `README.md`):
- `tiles/` — 1268 P3 coadd images + weights (detection layer)
- `exposures/` — 1180 unique single exposures, image+weight+flag (measurement layer; downloading)

So the run can consume everything off local disk — no CADC calls needed during processing.

## Setup tasks

- [ ] **Container.** `apptainer pull` the runtime image (build from `develop`, *not* the stale v1.0.1 tag), caching off `$HOME`:
  ```bash
  export APPTAINER_CACHEDIR=/project/def-mjhudson/unions-wl/.apptainer
  apptainer pull /project/def-mjhudson/unions-wl/shapepipe-runtime.sif \
      docker://ghcr.io/cosmostat/shapepipe:develop-runtime
  ```
- [ ] **Clone** `shapepipe` (`develop`) for the `example/cfis/` configs + `auxdir/`.
- [ ] **Point ShapePipe at local data.** In `config_tile_Git_vos.ini` and `config_exp_Gie_vos.ini`, set `RETRIEVE = symlink` and `INPUT_PATH` → `/project/def-mjhudson/unions-wl/{tiles,exposures}`. Verify `INPUT_FILE_PATTERN`/`INPUT_FILE_EXT` match our staged filenames (`CFIS.XXX.YYY.r.fits` / `.r.weight.fits.fz`; exposures `NNNNNNNp.fits.fz` + `.weight`/`.flag`).
- [ ] **Tile list.** Write `tile_numbers.txt` with the 2–4 chosen tiles.
[...]

PR #823, opened 2026-07-14
fix(setools): guard _mode against degenerate zoom window

cailmdaley:
## Problem
During the Nibi test run in #808, several single-exposure CCDs crashed in the SEtools star-selection step with `ValueError: Too many bins for data range. Cannot create 70 finite-sized bins.` The affected CCDs have FWHM distributions where the stellar-locus zoom window collapses onto a cluster of near-identical float32 values, so `np.histogram` is asked to build 70 bins over an (effectively) zero-width range.

## Root cause
`SETools._mode` iteratively zooms into the histogram peak. When the zoomed data are degenerate (all values equal to float32 precision), the bin-edge computation produces a zero/denormal bin width and numpy raises before a mode can be returned.

## Fix
Guard the zoom loop in `shapepipe/modules/setools_package/str_handler.py`: when the zoom window becomes degenerate, return the (single) value directly instead of histogramming it. +18/-1 lines, no behaviour change for healthy inputs.

## Evidence / testing
- Standalone repro in the runtime container: a degenerate near-identical-float32 cluster returns mode 4.0 on this branch; the identical input against `origin/develop`'s `str_handler` raises the 70-bins `ValueError`.
- Healthy control input returns 4.0798, bit-for-bit identical to develop.
- Full setools rerun in the container on a previously-crashing CCD completes and selects 56 stars.

### DIFF
--- a/src/shapepipe/pipeline/str_handler.py
+++ b/src/shapepipe/pipeline/str_handler.py
@@ -290,6 +290,19 @@ class StrInterpreter(object):
 
         iteration = 0
         while diff > eps:
+            if len(data) == 0:
+                raise ValueError(
+                    "Mode computation failed: zoom window emptied out"
+                    f" after {iteration} iterations"
+                )
+            if np.ptp(data) <= eps:
+                # The zoom window has degenerated: the remaining values span
+                # less than the requested accuracy (e.g. near-identical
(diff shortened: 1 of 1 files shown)


martinkilbinger (approved):
nice fix

Issue #662, opened 2023-11-08
Extend size of bright star masks [BUG]


lbaumo:
<img width="296" alt="Screen Shot 2023-10-27 at 1 42 19 PM" src="https://github.com/CosmoStat/shapepipe/assets/20316516/018c74bb-9906-4944-87a0-fc81eadd851e">
**Describe the bug**
Our masks for saturated stars are not big enough, we are getting object contamination in our catalog.

cailmdaley:
Possible major contributor found while debugging a first P3 test run on Nibi (#808): `mask.py` averages the four GSC magnitudes with `is not None` guards that let NaN through, so any bright star with missing GSC bands gets `mean = NaN`, fails `NaN < HALO_MAG_LIM`, and silently receives **no halo mask at all** — and the brightest (saturated) stars are precisely the ones most likely to have missing bands. On one test CCD, fixing this took the halo-star count from 57 to 118, the recovered objects being the brightest in the field (mag 5.7–9.5).

Fix is up in #824. If contamination persists after that, the remainder would point at halo *radius* scaling rather than missing masks — we plan to quantify (detections in halo annuli before/after) on a 10-tile P3 run.

— Claude (Fable) on behalf of Cail.

martinkilbinger:
After discussion with Mike and @aguinot it would be a good idea to have consistent and conservative masking within ShapePipe at exposure level and the structural masks / updated THELI setting that Hendrik is using.

PR #824, opened 2026-07-14
fix(mask): exclude NaN GSC magnitudes from halo/spike mean magnitude

cailmdaley:
## Problem
During the Nibi test run in #808, the mask module produced halo/spike masks that silently omitted some of the brightest objects in the field. The GSC catalogue used to seed halo masks contains entries with NaN magnitudes; these poisoned the mean-magnitude computation used to scale halo sizes, so bright stars near NaN entries were mishandled and dropped from the mask.

## Root cause
`mask.py` computes a mean over GSC magnitudes without excluding NaNs, so any NaN entry propagates and corrupts the halo-magnitude logic.

## Fix
Exclude NaN GSC magnitudes from the halo/spike mean-magnitude computation in `shapepipe/modules/mask_package/mask.py` (+18/-19 lines vs develop).

## Evidence / testing
- On the affected Nibi tile, the halo star count goes from 57 to 118, now including very bright objects in the mag 5.7–9.5 range; all 57 originally-selected halo stars are unchanged.
- End-to-end mask rerun on exposure 2586338-11 completes with the corrected halo set.
- Honesty note: the poster-child "bright object" recovered here is actually the galaxy NGC 5949, shredded by GSC into a Class-0 entry — the fix is still correct (NaN magnitudes must not poison the mean), but that particular object is a catalogue artefact rather than a bright star.

---

**Related: #662** — bright stars with missing GSC bands silently received *no* halo mask at all, which presents exactly as the under-masked bright stars reported there (on the test CCD, the halo-star count went 57 → 118 with this fix, and the recovered objects are the brightest in the field, mag 5.7–9.5). Worth re-assessing #662 once this is in; any residual contamination would then point at halo *radius* scaling rather than missing masks.

### DIFF
--- a/src/shapepipe/modules/mask_package/mask.py
+++ b/src/shapepipe/modules/mask_package/mask.py
@@ -888,26 +888,25 @@ class Mask(object):
         star_zip = zip(*(stars[k] for k in keys_to_use))
 
         for ra, dec, Fmag, Jmag, Vmag, Nmag, clas in star_zip:
-            mag = 0.0
-            idx = 0.0
-
-            # Compute mean magnitude
-            if Fmag is not None:
-                mag += Fmag
-                idx += 1.0
-            if Jmag is not None:
-                mag += Jmag
-                idx += 1.0
-            if Vmag is not None:
-                mag += Vmag
-                idx += 1.0
-            if Nmag is not None:
-                mag += Nmag
-                idx += 1.0
-            if idx == 0.0:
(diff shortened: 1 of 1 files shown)


martinkilbinger (approved):
Another bug that might have been introduced when transitioning from shell command findgsc to python astroquery. Also, luckily or accidentally not severe due to structural masks in post-processing.

PR #825, opened 2026-07-14
fix(sextractor): tolerate WCS inversion divergence in post-process

cailmdaley:
## Problem
During the Nibi test run in #808, the SExtractor post-processing step crashed on some tiles with astropy `NoConvergence` errors from inverse-WCS evaluation, while the same configuration had previously run cleanly. The regression appeared with the astropy 6 → 7 pin bump: astropy 7's epoch-matching changed which exposure footprints are evaluated, feeding positions far outside a CCD's valid WCS region into `world_to_pix`, where the iterative inversion diverges.

## Root cause
Post-processing tests every object position against every exposure WCS with no footprint prefilter and no divergence tolerance; positions far off-CCD make the SIP inversion diverge and raise.

## Fix
In `sextractor_script.py` / `sextractor_runner.py` (+95/-12 vs develop): prefilter objects against each exposure footprint before inversion, and tolerate residual divergence (`quiet=True` handling) instead of crashing, so far-off-CCD positions are simply counted as not-in-epoch.

## Evidence / testing
- A previously-succeeding tile reproduces bit-for-bit identical `N_EPOCH` values with the fix applied.
- The previously-crashing tile now completes post-processing for all 31,927 objects with zero divergence warnings.

### DIFF
--- a/src/shapepipe/modules/sextractor_package/sextractor_script.py
+++ b/src/shapepipe/modules/sextractor_package/sextractor_script.py
@@ -10,6 +10,7 @@ import re
 
 import numpy as np
 from astropy.io import fits
+from astropy.wcs.wcs import InvalidCoordinateError
 from sqlitedict import SqliteDict
 
 from shapepipe.pipeline import file_io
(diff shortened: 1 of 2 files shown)


martinkilbinger (approved):
Nice fix! I remember I saw this behaviour in the past but didn't know how to solve it.

PR #828, opened 2026-07-14
Inject controlled PSF-model error in make_data (testing)

cailmdaley:
`make_data` renders the galaxy with the true PSF and, by default, hands ngmix that same PSF as the model stamp — zero deconvolution error. Two optional knobs let the model stamp differ from the truth: `psf_model_fwhm_ratio` (size error) and `psf_model_shear` (shape error), so the digital twin can probe the one shape systematic metacal's response doesn't divide out.

Defaults reproduce the legacy output byte-for-byte; injected error perturbs only the PSF model stamp, never the galaxy image. Both pinned by tests.

— Claude (Opus) on behalf of Cail

### DIFF
--- a/src/shapepipe/testing/simulate.py
+++ b/src/shapepipe/testing/simulate.py
@@ -24,6 +24,8 @@ def make_data(
     psf_shear=(0.0, 0.0),
     wcs=None,
     return_centers=False,
+    psf_model_fwhm_ratio=1.0,
+    psf_model_shear=None,
 ):
     """Simulate an exponential galaxy with Moffat PSF.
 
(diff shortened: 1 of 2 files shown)


martinkilbinger:
@claude not sure I understand: The test asserts that shear and PSF are equal with and without perturbed PSF size and shape. Why whould these tests not fail? Would PSF size and shape residual (or is this not it?) not introduce additive and multiplicative shear biases?

cailmdaley:
yes exactly, this is less of a test-of-correctness than adding a knob to allow us to characterize PSF model error (e.g. what happens to the response when the PSF is mis-modelled in a controlled way).

PR #852, opened 2026-07-16
Snakemake orchestration for the real-data pipeline

cailmdaley:
Closes #848
Closes #873
Closed #844 

The PRD in #848 is the design reference; this PR realizes it.

**What this replaces:** the three-level bash orchestration of real-data runs (`run_job_sp_canfar_v2.0.bash` → `job_sp_canfar_v2.0.bash` → per-site sbatch wrappers). The bash path is left in place until this workflow is validated and adopted; removing it is a follow-up PR.

**Shape:** `workflow/` at top level — Snakefile + three rule files, six stdlib scripts, committed configs with fixed paths, one nibi profile. `workflow/bin/sp` is the entry point: `sp run` (two snakemake invocations, prepare then compute, joined by a sqlite tile↔exposure index), `sp report` (standalone, runs mid-flight), `sp container` (pull/status/sandbox/exec — per-user image cache, resolution sandbox → SIF → configured default), `sp cancel`.

**Additional changes not described in #848.** Three change scientific behavior for all users, bash path included:

- **ngmix RNG:** position-seeded per-object RNG is now the only mode. The tile-seeded mode is removed (results depended on how a tile was chunked); a config with `SEED_FROM_POSITION = False` now fails loudly instead of running the retired mode.
- **setools star split (#873):** `rand_split` is now seeded from the file number (previously unseeded), making the PSF train/validation split reproducible — and different from any previous draw.
- **`STAR_THRESH` 20 → 22** in the two `config_tile_PiViVi_canfar_*.ini`, aligning with the 80% convention already in the exposure PSF configs.

Mechanical enablers and hardening, no output changes: ngmix ID range / `RUN_NAME` read env-expanded; `merge_sep_cats` chunk-path derivation fixed for absolute paths with digits in parent dirs; Vizier responses no longer cached to `$HOME` (~50 GB at campaign scale); star catalogues written atomically; unit tests for the chunk partition, tile reclamation, and run report.

**Reviewing:** read the PRD's D1–D6 first; each rule file's header comments carry the local why. Run results and DR6 projections are in the benchmark comment below.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #873, opened 2026-08-29
fix(psf): deterministic star train/validation split; align tile STAR_THRESH with the 80% convention

cailmdaley:
Two small fixes to how the PSF star sample is built, found while auditing the scientific decisions embedded in the pipeline configs.

**1. Seed the train/validation split.** `SETools._make_rand_split` drew the 80/20 star split from unseeded `np.random`, so the star sample entering the PSF model — and everything downstream of it — differed between identical runs. The split is now a permutation seeded from the unit's file number: deterministic per CCD, independent of processing order, same philosophy as ngmix's `SEED_FROM_POSITION`. (This changes the realized split once — equivalent to one more random draw, now frozen.)

**2. STAR_THRESH 20 → 22 in the tile PSF-interp configs.** fdc86553 (2020) deliberately raised the per-CCD acceptance threshold to 22 to account for the 80% training split, but only in `config_exp_psfex.ini` (the validation pass). The tile configs that feed galaxies kept the stale 20, so the science path admits CCDs under the pre-split standard while validation gates at 22. We think the right fix is to align on 22.

Both change outputs slightly (a new frozen split; CCDs with 20–21 accepted stars now excluded, matching validation).

https://claude.ai/code/session_01Jrs8TeTCccRMC45QVGKPeY

### DIFF
--- a/example/cfis/config_tile_PiViVi_canfar_sx.ini
+++ b/example/cfis/config_tile_PiViVi_canfar_sx.ini
@@ -79,7 +79,7 @@ POSITION_PARAMS = XWIN_WORLD,YWIN_WORLD
 GET_SHAPES = True
 
 # Number of stars threshold
-STAR_THRESH = 20
+STAR_THRESH = 22
 
 # chi^2 threshold
 CHI2_THRESH = 2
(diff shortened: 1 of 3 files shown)


cailmdaley:
bundling into #852

PR #847, opened 2026-07-16
Consistent use of the external healsparse masks

cailmdaley:
Closes #846. Closes #837. Part of the #845 epic; builds on #852 and merges after it.

Until now, ShapePipe made its own masks: the `mask` module rasterized GSC star positions, Messier/NGC regions, and the instrument flags into a per-image flag file, and the star selection cut on that image. The mask force's external masks never entered the pipeline.

This PR removes all mask generation from ShapePipe. Two kinds of mask input remain, ingested differently.

1. **The instrument flags** delivered with each exposure (bad columns, saturation) — the only masks that touch pixels. They flow as before: SExtractor reads them per CCD and records them per object as `IMAFLAGS_ISO`; ngmix sees them in its stamp weights.
2. **The UNIONS healsparse masks**, fixed on the sky — never rasterized. Each map is queried once per object at its (RA, Dec) and the answer becomes a catalogue column: `make_cat` writes every bit, per band, into `MASK_<band>`, unfiltered. All selection on these columns happens downstream.

**PSF star selection** — the one decision that cannot be redone later — cuts only on the instrument flags: they mark corrupted measurements, stars whose own photometry and shape are unreliable. For everything the sky masks describe, the pipeline deliberately starts by relying on PSFEx's outlier rejection alone, and the first runs measure how well that holds. If it proves insufficient, there is machinery to pass any of the external masks — for example the star mask — to the PSF star selection, switched on in config.

**Shape measurement** uses no external mask: ngmix zero-weights instrument-flagged pixels and drops heavily flagged epochs, exactly as before. Whether an object in a halo, a trail, or a masked star enters the science sample is decided downstream from the `MASK_<band>` columns.

The survey footprint is a separate, positive coverage map — the exposure map of CCDs with a PSF model (#797), or the pointing-coverage map — and the mask bits subtract from it to give the final footprint.

## The mask products

Sky-fixed boolean healsparse maps, one per bit, nside 131072 ≈ 1.6″; current products `mask_ugriz_nside131072_n<bit>.hsp` (2026-08-05, post-GSC2-fix; locations on the [wiki](https://unions.skysurvey.cc/en/HostedScienceProjects/2025_2D_Cosmic_Shear/masking)):

| bit | value | meaning |
|---|---|---|
| 0, 1 | 1, 2 | faint / bright star halo |
| 2 | 4 | star mask (the star's core and spikes) |
| 3 | 8 | manual mask (large galaxies) |
| 4–8 | 16–256 | no *u g r i z* data |
[...]

Issue #844, opened 2026-07-16
Configurable retention of intermediate ShapePipe products


cailmdaley:
The Nibi orchestration keeps only the final catalogue and the star-catalogue inputs; everything else is deleted to stay inside scratch. Retention is a list of glob patterns in the config. From the 2026-09-08 call: add the PSF model (small; needed for image sims) and check that all SExtractor columns reach the final catalogue. Extend the keep-list as validation on the 64-tile run shows what's missing.

PR #879, opened 2026-09-01
Persist PSF products and add the two campaign-level merges (exp_persist, star_cat_merge, final_cat_merge)

cailmdaley:
Stacked on #852. No campaign so far has produced the two merged files sp_validation reads: `clean_exposure` reclaims the whole exposure store and only per-tile `final_cat` reaches `products_dir`. This adds the missing persistence and the two merges (@martinkilbinger's comment on #852).

- **`exp_persist`** (between `exp_psf` and `clean_exposure`): packs each exposure's PSF products into one tar, `<products_dir>/exp/<shard>/<exp>/psf/<exp>.tar`, plus a manifest. One tar because inodes bind on /project (~200 loose files per exposure, ~2 M at DR6). `validation_psf` — the star catalogue's input — is always packed (~2 MB/exposure). `persist_exp:` in config is the *retention* list on top: named products (`psf_model`, `psfex_cat`, `star_selection`, …; default `[psf_model]`, per #844 and the 2026-09-08 call), raw globs accepted, additive — a config change never drops a member already on disk. It rides on `params`, so editing it re-packs without re-running the PSF job.
- **`star_cat_merge`**: every exposure's `validation_psf` members, read straight out of the tars, → `<products_dir>/full_starcat_<campaign>.hdf5` (`exposures/<exp>` datasets, the 16 columns `merge_starcat_runner` emitted, native dtypes). This replaces the `combine_runs.bash psf` + `merge_starcat_runner` product. No (X,Y)→WCS pass — `validation_psf` carries RA/DEC.
- **`final_cat_merge`**: all tiles' `final_cat-<ID>.fits` → `<products_dir>/final_cat_<campaign>.hdf5` (sp_validation's `galaxy_cat_path`), via `create_final_cat.py`'s column extraction. Keeps sp_validation's `patches/<name>/<ID>` layout with the campaign as the single group — legacy schema only; ShapePipe v2 has no patches (CosmoStat/sp_validation#340).

**Notes.** Both merges *reconcile* (shared `hdf5_reconcile.py`): an appended unit reads only itself, a dropped unit's dataset is removed, a column-set change refreshes everything, a no-op leaves the file untouched; memory is flat in the campaign. Both derive their input set from `tile_list` + `run_index.sqlite`, identical to the DAG's, and fingerprint unit IDs on `params` — no paths on argv (128 KiB limit). An exposure counts as reclaimed on a tombstone *or* a persisted manifest with no scratch `exp_psf` manifest, so a /scratch purge, `clean: false`, or a pre-`exp_persist` cleanup never triggers a VOS rebuild (measured on smk-g6: 0 exposure jobs in all three). Memory requests scale with input size, capped at `max_mem_mb`. Upstream fixes along the way: `create_final_cat.py` no longer writes uninitialised memory into non-param columns and orders columns by the param file; `MergeStarCat*` accumulate arrays (two-pass for PSFEx) instead of Python lists and handle optional columns per file; `final_cat.param` drops `IMAFLAGS_ISO` (never produced tile-side since #847) and renames `NGMIX_MOM_FAIL` → `NGMIX_MCAL_TYPES_FAIL`.

**Verified.** `exp_persist` end-to-end on smk-m2 (127/127 exposures); tar form, both merges and the reconcile paths on fixtures (byte-stable, add/refresh/remove/no-op, column values equal to the FITS reference); `final_cat_merge` over smk-g6's 64 real catalogues (2.5 GB in 20 s, 151 MB RSS) once `NGMIX_NEIGHBOUR_FLAG` is excluded — g6 predates it; dry-run on smk-g6 (`final_cat_merge` 1 job; `star_cat_merge` 0 — its exposures were reclaimed before `exp_persist` existed). First real run: the next 64-tile campaign.

`final_cat_merge` and `star_cat_merge` stamp `sp run`'s code snapshot (head, branch, dirty flag and dirty file list, snapshot time) onto the merged HDF5 root attributes, so campaign products carry the same code provenance the run directory records.

https://claude.ai/code/session_01QbnPCyzuDNTgkg715pHhar

### DIFF
--- a/scripts/python/create_final_cat.py
+++ b/scripts/python/create_final_cat.py
@@ -338,13 +338,19 @@ def copy_data(param_list, extracted_data, dtype):
     """Copy Data.
 
     """
-    # THE REQUESTED COLUMNS ONLY, in the SOURCE catalogue's order. Allocating
-    # with the source's full dtype and filling only the requested columns left
-    # every other column as uninitialised memory: meaningless values in the
-    # output file, and different bytes on every run of this tool over the same
-    # inputs. The parameter file says which columns the merged catalogue is
-    # for; those are the columns it gets.
-    columns = [col for col in (dtype.names or ()) if col in set(param_list)]
+    # THE REQUESTED COLUMNS ONLY, IN THE PARAMETER FILE'S ORDER. Two things
+    # are being fixed here and they are easy to conflate. Allocating with the
+    # source's full dtype and filling only the requested columns left every
+    # other column as uninitialised memory — meaningless values, and different
+    # bytes on every run over the same inputs. And ordering the result by the
+    # SOURCE catalogue's columns made the output dtype a property of the
+    # catalogue rather than of the parameter file: two tiles written by
+    # different ShapePipe versions, whose catalogues order or extend their
+    # columns differently, then landed in one merged file with two different
+    # structured dtypes, which np.concatenate refuses. The parameter file is
+    # the schema; it says which columns AND in what order.
+    wanted = set(dtype.names or ())
+    columns = [col for col in param_list if col in wanted]
     subset = np.dtype([(col, dtype[col]) for col in columns])
 
     # Initialize new data structure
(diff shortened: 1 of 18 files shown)


cailmdaley:
transferring @martinkilbinger's comment on star/PSF products here:
https://github.com/CosmoStat/shapepipe/pull/852#issuecomment-5602122537

